PR #14, opened 2022-11-09
Filter which providers should be available to users

### LINKED ISSUE #4: Limit available providers
We should choose which providers we want to enable, and disable all of the others.

We'll probably want accounts with greater access to have more strict requirements. e.g., Core committers should use webauthn as primary and TOTP as backup, while regular users can can use TOTP as primary and (maybe) email as backup.

iandunn:
Fixes #4 

I went with the simpler approach discussed in that issue, but am open to other ideas

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -23,6 +23,7 @@ require_once $_tests_dir . '/includes/functions.php';
  * Manually load the plugin being tested.
  */
 function _manually_load_plugin() {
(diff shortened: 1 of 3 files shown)


dd32 (approved):
Approach looks good to me, to fix #4.

Confirmed (via wpdirectory.net) that seemingly no other two-factor extenders use a priority >10 for the filter (and have no reason to), so at 99 should be safe here.

Issue #6, opened 2022-11-03
Enforce 2FA for privileged accounts


iandunn:
Core/Meta committers, super-admins, project leaders, committers to plugins w/ > 50k active installs, etc should be required to use 2FA of some kind. How strict we are should depend on what kind of access they have.

- [x] Require for super admins
- [x] Require for WordCamp super deputies
- [x] Require for WordCamp subroles
- [x] Disable on production for beta testing
- [ ] Enable on production once launches for all users
- [ ] (optional) Reorganize `capes.php` to make it easier to work with
- [ ] Determine which other roles should require it
- [ ] Require it for those roles
- [ ] Explore the idea of encouraging (but not requiring) it for some roles - probably open a separate ticket for that

Related: 
* #4
* https://github.com/WordPress/two-factor/issues/255
* https://github.com/WordPress/two-factor/pull/239
* https://github.com/WordPress/two-factor/issues/249

iandunn:
Related: `wpcom_vip_two_factor_filter_caps()` from https://github.com/Automattic/vip-go-mu-plugins/blob/develop/two-factor.php

iandunn:
Rough idea:

* Default is opt-in
* Encouraged if
    * Make team representative
    * Admin on any w.org/wordcamp.org site
* Mandatory with TOTP or hardware key+webauthn if:
    * has write access to a github.com/wordpress repo
    * plugin has more than X installs
    * theme has more than Y installs
    * wordcamp trusted deputy
* Mandatory using hardware key + webauthn if:
    * has write access to (core|build|meta|dotorg|wordcamp|code).svn.w.org
    * w.org super admin
    * admin access to any github.com/wordpess repo
    * any other form of leadership – josepha, matt, chloe, matias, etc

Requiring a hardware key for the highest privilege group would prevent phishing. Most folks will never buy one though, so it’s important to not push it too hard. Maybe we give frequent contributors a coupon code so they can buy one on mercantile.w.org for the cost of shipping, or half price or something.

iandunn:
To facilitate this, we may want to reorganize `capes.php`, so it has a group of functions like

 can have list of all privledges users statically. so someone is listed as a supe even if they're not proxied. still only assigned supe if proxied, but can see 

etc

then at the end

iandunn:
[...]

PR #17, opened 2022-11-17
Enforce 2FA for super admins

iandunn:
See #6 

This adds a few roles to get started. If this looks good, we can add more in a future PR.

This will require a small change to `capes.php`:

```diff
-	$super_admins =& $supes;
+	// This is a copy rather than a reference so that `$supes` can represent folks who _might_ have elevated
+	// privileges, and `$super_admins` can represent folks who _do_ currently have them.
+	// `wporg-two-factor` needs that distinction to disable capabilities until 2FA is setup.
+	$super_admins = array_merge( $supes );
```

### DIFF
--- a/wporg-two-factor.php
+++ b/wporg-two-factor.php
@@ -10,6 +10,9 @@
  */
 
 namespace WordPressdotorg\Two_Factor;
+use Two_Factor_Core;
+use WP_User;
+
 defined( 'WPINC' ) || die();
 
 // Disabled until ready for launch.
@@ -18,7 +21,10 @@ if ( 'production' === wp_get_environment_type() ) {
 }
 
 
-add_filter( 'two_factor_providers', __NAMESPACE__ . '\two_factor_providers', 99 ); // Must run after all other plugins.
+add_filter( 'two_factor_providers', __NAMESPACE__ . '\two_factor_providers', 99 ); // Must run _after_ all other plugins.
+add_action( 'set_current_user', __NAMESPACE__ . '\remove_super_admins_until_2fa_enabled', 1 ); // Must run _before_ all other plugins.
+add_action( 'login_redirect', __NAMESPACE__ . '\redirect_to_2fa_settings', 105, 3 ); // After `wporg_remember_where_user_came_from_redirect()`, before `WP_WPorg_SSO::redirect_to_policy_update()`.
+add_action( 'user_has_cap', __NAMESPACE__ . '\remove_capabilities_until_2fa_enabled', 99, 4 ); // Must run _after_ all other plugins.
 
 /**
  * Determine which providers should be available to users.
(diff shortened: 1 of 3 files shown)


wporg-two-factor.php
```diff
@@ -33,3 +38,114 @@ function two_factor_providers( array $providers ) : array {
...
+	}
+
+	return $required;
+}
+
+/**
+ * Redirect a user to their 2FA settings if they need to enable it.
+ *
+ * This isn't usually necessary, since WordPress will prevent Subscribers from visiting other Core screens, but
+ * sometimes plugins add screens that are available to Subscribers (either intentionally or not).
```
dd32, line 116: Given they have no caps, I'm not sure this is really something that needs to be concerned with.

If a admin page/panel is enabled for everyone, including subscribers, that's hardly something we need to enforce 2FA for a super-admin for.

To me this just feels like it's going to result in some kind of buggy behaviour at some point.. 

I would personally implement this as the cap-removal, and a post-login error page / forced-redirect to profile with the "You must enable 2FA" message.

Not saying we should remove this, just suggesting that it might not strictly be needed.

wporg-two-factor.php
```diff
@@ -33,3 +38,114 @@ function two_factor_providers( array $providers ) : array {
...
+	if ( 0 === $user->ID || ! user_requires_2fa( $user ) ) {
+		return $allcaps;
+	}
+
+	if ( ! Two_Factor_Core::is_user_using_two_factor( $user->ID ) ) {
+		// This also relies on `remove_super_admins_until_2fa_enabled()`, see notes in that function.
+		$allcaps = array(
+			'subscriber' => true,
+			'read'       => true,
+		);
```
dd32, line 83: Another option here I think is to return the `do_not_allow` cap when the requested cap is anything other than the whitelisted caps.

I believe that would override the super-admin always-truthful caps too, but `remove_super_admins_until_2fa_enabled()` would be best kept still then too.

dd32 (approved):
LGTM, I've added some comments on my thoughts of implementation, but I think this is good to go regardless of those.

wporg-two-factor.php
iandunn: 🤔 So you're saying the difference is essentially:

```diff
- Login > redirect to profile > try to visit another page > redirect to profile
+ Login > redirect to profile > try to visit another page > let them access it
```

? That seems fine to me 👍🏻 I'm more concerned about the UX, and agree that viewing pages available to subscribers isn't a big deal.

wporg-two-factor.php
iandunn: Ah, that's a good point 👍🏻

iandunn: Er, nevermind, it looks like `user_has_cap` doesn't even get called in that case.

https://github.com/WordPress/wordpress-develop/blob/a71d7373ad6ce0a8f252717832f42aebc3e08854/src/wp-includes/class-wp-user.php#L780-L809

wporg-two-factor.php
iandunn: Changed in 55de8ab29046e54c2091e31bfee35758f21a7515

iandunn:
I'll wait until Monday to deploy the `capes.php` change, and a few related ones I noticed since then. I'll go ahead and merge this now though, since I think it's ready.

Issue #21, opened 2022-11-29
Only allow backup codes when another 2FA method exists


pkevan:
Currently you can setup backup codes without an additional method (TOTP, auth key). 

If we want to block this workflow, and tie backup codes to only be used as an additional method, then we'll need to rework the logic and possibly restrict backup codes being a primary method.

pkevan:
Related: #15

pkevan:
See: https://github.com/pkevan/two-factor/commit/0f6ee66d3c2136a430d2d37655b432795f70748e

iandunn:
Related #23 

> See: https://github.com/pkevan/two-factor/commit/0f6ee66d3c2136a430d2d37655b432795f70748e

Was that meant to be posted to #22?

If we're writing a custom UI from scratch, I wonder if it'd be better to use React and Gutenberg components/utilities, instead of jQuery/etc?

pkevan:
> Was that meant to be posted to https://github.com/WordPress/wporg-two-factor/issues/22?

Yes! Good catch :)

pkevan:
> If we're writing a custom UI from scratch, I wonder if it'd be better to use React and Gutenberg components/utilities, instead of jQuery/etc?

Probably, i'm just prototyping the functionality and following pre-existing patterns in the plugin.

pkevan:
We can probably achieve this using the `two_factor_providers` filter.

pkevan:
https://github.com/WordPress/wporg-two-factor/pull/24 as a potential method for doing this.

TobiasBg:
I only found this repo after opening a ticket in Meta Trac, but for reference: https://meta.trac.wordpress.org/ticket/6825

iandunn:
Hey 👋🏻 , sorry you experienced that. https://github.com/WordPress/two-factor/issues/507 is the main cause for it, but #75 and #47 will help mitigate it in the mean time.

TobiasBg:
Hi @iandunn, thanks for the context! Yes, that sounds exactly like what happened. The fix sounds reasonable to me as well. (I'm all set again already, thanks to a forums moderator.)

outdoor2kode:
I guess we can close this issue for now since #75 has already been merged and #47 can stand on its own.
@iandunn @pkevan  If you feel it's necessary, please feel free to reopen it.

Issue #18, opened 2022-11-23
Create UI


pkevan:
Pulling out the discussion from slack and including the screenshots for reference. All text pulled as is, so not necessarily me saying it 😄 

https://a8c.slack.com/archives/C0393K4ADM3/p1668679184861899?thread_ts=1668663080.269429&cid=C0393K4ADM3
So to summarize my advice, and I’m sure there are a lot of technical gotchas that make this a unique challenge:
- Unify profiles if you can, otherwise place the flow where it makes the most sense. My instinct is the main profile, but if your instinct says forum profile, alright then.
- Instead of a single box that sits on the main profile, see if it can be a single “activate” and “deactivate” button, that moves you into a stepped flow instead of all happening in the same place.

### Example of the potential flow:
**the [WP.org](http://wp.org/) profile could have a similar single button under a “Security” subheading: “enable 2fa”.**
<img width="1392" alt="1" src="https://user-images.githubusercontent.com/2290623/203520646-d4431004-ef59-43f9-9306-d0bd910286da.png">

**Clicking said “enable 2fa” button would take you to an entirely new page, step 1 of a sequence of steps, and the first step could be choosing an app or sms, depending on the tech available**
<img width="1392" alt="Screenshot 2022-11-17 at 10 53 47" src="https://user-images.githubusercontent.com/2290623/203520811-a2bd59d3-d731-4d94-9831-5f65d0de29d8.png">

**That would then show the QR code contextually to the step**
<img width="1392" alt="Screenshot 2022-11-17 at 10 54 33" src="https://user-images.githubusercontent.com/2290623/203520917-01bbca52-e7b8-40c6-980f-4ae883b7e613.png">

**Activating would let you see backup codes, and you can check the box to enable the next step**
<img width="1392" alt="Screenshot 2022-11-17 at 10 55 24" src="https://user-images.githubusercontent.com/2290623/203521018-582026dc-72d1-4a2e-9746-f8969e2da3c0.png">

**Finally, a nice little notice that you’ve activated 2fa, and the “Activate 2fa” button on the security section of your profile would instead become “disable 2fa”.**
<img width="1392" alt="Screenshot 2022-11-17 at 10 56 00" src="https://user-images.githubusercontent.com/2290623/203521056-343f55ca-b481-4597-a095-7e69c62242bd.png">

pkevan:
In follow-up to #5

tellyworth:
[...]

PR #24, opened 2022-11-30
Check user has a desired provider set

pkevan:
Disallow backup codes unless a desired provider exists

### DIFF
--- a/wporg-two-factor.php
+++ b/wporg-two-factor.php
@@ -35,8 +35,13 @@ function two_factor_providers( array $providers ) : array {
 	$desired_providers = array(
 		'Two_Factor_WebAuthn'     => '',
 		'Two_Factor_Totp'         => '',
-		'Two_Factor_Backup_Codes' => '',
 	);
+
+	// Check current user has a primary method setup, before allowing backup codes
+	$user_providers = Two_Factor_Core::get_enabled_providers_for_user();
+	if ( ! empty( array_intersect_key( $user_providers, $desired_providers ) ) ) {
+		$desired_providers['Two_Factor_Backup_Codes'] = '';
+	}
 
 	return array_intersect_key( $providers, $desired_providers );
 }


wporg-two-factor.php
```diff
@@ -35,9 +35,14 @@ function two_factor_providers( array $providers ) : array {
 	$desired_providers = array(
 		'Two_Factor_WebAuthn'     => '',
 		'Two_Factor_Totp'         => '',
-		'Two_Factor_Backup_Codes' => '',
```
iandunn, line 38: Will this cause any problems w/ the flow in #18? The provider [wouldn't be loaded](https://github.com/WordPress/two-factor/blob/555d88685087c0f9e08030cf93bdacfb749dbded/class-two-factor-core.php#L133-L151) when the user visits the page, so we might not be able to generate the backup codes during activation, unless we explicitly `require` it in the Ajax callback.

The `two_factor_enabled_providers_for_user` might solve any issues there, and seems a bit more appropriate.

wporg-two-factor.php
```diff
@@ -35,9 +35,14 @@ function two_factor_providers( array $providers ) : array {
...
 	);
 
+	// Check current user has a primary method setup, before allowing backup codes
+	$user_providers = Two_Factor_Core::get_enabled_providers_for_user();
+	if ( ! empty( array_intersect_key( $user_providers, $desired_providers ) ) ) {
+		$desired_providers['Two_Factor_Backup_Codes'] = '';
+	}
+
```
iandunn, line 45: Do you mind adding some basic tests to the suite? `composer run test:watch` and `test:coverage`

iandunn (approved):
Left a few questions/suggestions but they're not blockers 👍🏻

wporg-two-factor.php
pkevan: Possibly - although for the activation flow we could add in the backup codes through hooking in later to `two_factor_providers` since we'd be manipulating the standard flow anyway.

pkevan: Just revisited this, and not sure if either are a good option - if we remove the option to allow backup codes until totp or webathn are enabled, we'd have to re-trigger the code to allow using this method. If we use `two_factor_enabled_providers_for_user` do we want to disable a method that could potentially lock someone out, for example they try to reenable a method and it fails?

iandunn: 🤔 yeah, fair point. what problem is this trying to solve? maybe there's another way to do it?

pkevan: > what problem is this trying to solve?

In #21 we were trying to avoid users setting up backup codes without another preferred method, but maybe it's better just to prompt them to do so, rather than force them to using code?

The plugin does have logic that if you only have one provider it becomes the primary, so potentially in this scenario should there be an error in setup for a preferred provider the user could become unnecessarily locked out.

wporg-two-factor.php
```diff
@@ -35,9 +35,14 @@ function two_factor_providers( array $providers ) : array {
 	$desired_providers = array(
 		'Two_Factor_WebAuthn'     => '',
 		'Two_Factor_Totp'         => '',
-		'Two_Factor_Backup_Codes' => '',
 	);
 
+	// Check current user has a primary method setup, before allowing backup codes
+	$user_providers = Two_Factor_Core::get_enabled_providers_for_user();
```
iandunn, line 41:
```suggestion
	$user_providers = Two_Factor_Core::get_available_providers_for_user();
```

i didn't realize that this until recently, but the enabled/available functions are confusingly named. A provider can be "enabled", but not it's "available" until it's been _configured_. I think in this context we want the latter.

iandunn, line 41: Should we pass `bbp_get_displayed_user_id()` to the function, for the the situations where an admin is editing another user?

wporg-two-factor.php
iandunn: Would the ["enabled" vs "available" distinction](https://github.com/WordPress/wporg-two-factor/pull/24#discussion_r1102046729) might be useful in avoiding locking the user out?

iandunn:
I think this was superseded by #75, but please reopen if there's anything from it that we still want to merge.

PR #25, opened 2022-12-07
Block sending authentication cookies prior to 2FA being validated

dd32:
See the upstream https://github.com/WordPress/two-factor/issues/406 and https://github.com/WordPress/two-factor/pull/502

This should unblock that issue for WordPress.org.


iandunn (approved):
LGTM, but also looks like #502 is ready for merge, so we may not need this.

iandunn:
Closing since https://github.com/WordPress/two-factor/pull/502/ was merged

Issue #28, opened 2022-12-13
UI: Email


iandunn:
Implement the email field in the custom UI, see #18 for rough mockups

dd32:
Noting that I've opened  https://core.trac.wordpress.org/ticket/57413 to move the "confirm your email" code added in #32 upstream to core.

iandunn:
Nice 👍🏻 

Closing this now that #32 is merged

PR #32, opened 2022-12-21
WIP of the Email settings screen

dd32:
See #28

### DIFF
--- a/settings/rest-api.php
+++ b/settings/rest-api.php
@@ -1,0 +1,125 @@
+<?php
+
+namespace WordPressdotorg\Two_Factor;
+
+defined( 'WPINC' ) || die();
+
+add_action( 'rest_api_init', __NAMESPACE__ . '\register_user_fields' );
+add_filter( 'rest_pre_insert_user', __NAMESPACE__ . '\require_email_confirmation', 10, 2 );
+
+/**
+ * Register/Output some REST-API calls to be pre-loaded.
+ *
+ * This prevents the browser having to make the HTTP call before the react UI can be ready.
+ * This duplicates block_editor_rest_api_preload() as there is no generic function for performing this preloading.
+ * WARNING: This will output Javascript immediately if called during the page load if the wp-api-fetch script has already been output.
+ *
+ * @param array $preload_paths The REST API paths to be pre-loaded, must include prefixed slash.
+ * @return void
+ */
+function preload_api_requests( array $preload_paths ) : void {
+	global $post, $wp_scripts, $wp_styles;
+
+	/*
+	 * Ensure the global $post, $wp_scripts, and $wp_styles remain the same after
+	 * API data is preloaded.
+	 * Because API preloading can call the_content and other filters, plugins
+	 * can unexpectedly modify the global $post or enqueue assets which are not
+	 * intended for the block editor.
+	 */
+	$backup_global_post = ! empty( $post ) ? clone $post : $post;
+	$backup_wp_scripts  = ! empty( $wp_scripts ) ? clone $wp_scripts : $wp_scripts;
+	$backup_wp_styles   = ! empty( $wp_styles ) ? clone $wp_styles : $wp_styles;
+
+	$preload_data = array_reduce(
+		$preload_paths,
+		'rest_preload_api_request',
+		[]
+	);
+
+	// Restore the global $post, $wp_scripts, and $wp_styles as they were before API preloading.
+	$post       = $backup_global_post;
+	$wp_scripts = $backup_wp_scripts;
+	$wp_styles  = $backup_wp_styles;
+
+	$preload_js = sprintf(
+		'wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( %s ) );',
+		wp_json_encode( $preload_data )
+	);
+
+	// The script may have already been printed. In that case, add the pre-loading inline now.
+	if ( wp_script_is( 'wp-api-fetch', 'done' ) ) {
+		printf(
+			'<script>%s</script>',
+			$preload_js
+		);
+	} else {
+		wp_enqueue_script( 'wp-api-fetch' );
+		wp_add_inline_script(
+			'wp-api-fetch',
+			$preload_js,
+			'after'
+		);
+	}
+}
+
+/**
+ * Register any user meta that needs to be exposed.
+ */
+function register_user_fields(): void {
+	// Expose the `_new_email` user meta through the rest api as a custom user field.
+	// This is for "The user has a pending email change"
+	register_rest_field(
+		'user',
+		'pending_email',
+		[
+			'get_callback' => function( $user ) {
+				return get_user_meta( $user['id'], '_new_email', true )['newemail'] ?? false;
+			},
+			'update_callback' => function( $value, $user ) {
+				if ( '' === $value ) {
+					delete_user_meta( $user->ID, '_new_email' );
+					return true;
+				}
+			},
+			'schema' => [
+				'type'    => 'string',
+				'context' => [ 'edit' ],
+			]
+		]
+	);
+
+}
+
+/**
+ * Implement the "Require email confirmation" functionality for the REST API.
+ *
+ * TODO: This is a core bug. This should be handled by core.
+ * TODO: This should be moved to a WordPress.org mu-plugin.
+ * TODO: This generates urls to /support/wp-admin/profile.php?newuseremail=%s
+ *       bbPress also implements this functionality, through bbp_edit_user_email_send_notification()
+ *
+ * @param array $insert_data The user data being updated.
+ * @return array
+ */
+function require_email_confirmation( $insert_data, $request ) {
+	global $errors;
+
+	if ( isset( $insert_data->user_email ) ) {
+		$post_backup = $_POST;
+
+		// The POST fields needed by send_confirmation_on_profile_email().
+		$_POST['user_id'] = $insert_data->ID;
+		$_POST['email']   = $insert_data->user_email;
+
+		send_confirmation_on_profile_email();
+
+		if ( $_POST['email'] !== $insert_data->user_email || $errors->has_errors() ) {
+			$insert_data->user_email = $_POST['email'];
+		}
+
+		$_POST = $post_backup;
+	}
+
+	return $insert_data;
+}
--- a/settings/src/components/email-address.js
+++ b/settings/src/components/email-address.js
@@ -4,6 +4,8 @@
  */
 import { Button, TextControl } from '@wordpress/components';
 import { useState } from '@wordpress/element';
+import { useSelect } from '@wordpress/data';
+import { saveUser, saveEntityRecord } from '@wordpress/core-data';
 
 /**
  * Render the Email setting.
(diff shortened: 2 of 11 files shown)


settings/rest-api.php
```diff
@@ -0,0 +1,125 @@
...
+	);
+
+}
+
+/**
+ * Implement the "Require email confirmation" functionality for the REST API.
+ *
+ * TODO: This is a core bug. This should be handled by core.
+ * TODO: This should be moved to a WordPress.org mu-plugin.
+ * TODO: This generates urls to /support/wp-admin/profile.php?newuseremail=%s
```
iandunn, line 99:
```suggestion
 * TODO: This generates urls to /support/wp-admin/profile.php?newuseremail=%s, but we want `get_edit_account_url()` instead
```

I missed that it was going to `wp-admin` at first glance, so explicitly saying what we want instead makes it more obvious

settings/src/script.js
```diff
@@ -66,9 +71,19 @@ function Main() {
...
+
+		// Reset to initial after navigating away from a page.
+		if ( userDataNeedsReset ) {
+			editUserData( userData );
+		}
+
 		setScreen( screen );
 	}
 
+	if ( ! userData ) {
```
iandunn: Using the `isResolving` bool or `status === RESOLVING` might be a bit safer if there are any functional differences between those and checking `! record`, but I have no idea if there actually are any.

I guess `status` could also be `ERROR`, but it may not be worth handling that.

settings/src/script.js
```diff
@@ -81,24 +96,12 @@ function Main() {
 			}
 
 			<div className={ 'wporg-2fa__' + screen }>
-				<CurrentScreen clickScreenLink={ clickScreenLink } />
+				<CurrentScreen clickScreenLink={ clickScreenLink } userId={ userId } userData={ userData } />
```
iandunn:
```suggestion
				<CurrentScreen clickScreenLink={ clickScreenLink } userData={ userData } />
```

This doesn't seem necessary, since we're also passing `userData.id`. Not a big deal, though.

`EmailAddress` would need to be updated to use `userData.id` if this is accepted.

settings/src/components/email-address.js
```diff
@@ -0,0 +1,94 @@
...
+		}
+	};
+
+	if ( ! record ) {
+		return <Spinner />
+	}
+
+	return (
+		<>
+			{ record.pending_email && ! justChangedEmail && <Notice status="warning" className="actions-on-right" isDismissible={ false } actions={ [ { label: "Cancel change", onClick: handleDiscard } ] }>
```
iandunn: The length of this, plus the way it's combining logic and markup, makes it relatively hard to read IMO. Something like this feels much easier to me:

```jsx
{ record.pending_email && ! justChangedEmail && 
	<Notice 
		status="warning"
		className="actions-on-right" 
		isDismissible={ false } 
		actions={ [ { label: "Cancel change", onClick: handleDiscard } ] }
	>
		<p>
			There is a pending email change to { record.pending_email }.<br />
			Please check your email for a confirmation link.
		</p>
	</Notice>
}
```

Same for the one below.

settings/src/components/email-address.js
```diff
@@ -0,0 +1,94 @@
+
+/**
+ * WordPress dependencies
+ */
+import { Button, TextControl, Notice, Spinner } from '@wordpress/components';
+import { useState, useEffect } from '@wordpress/element';
```
iandunn:
```suggestion
import { useState } from '@wordpress/element';
```

I don't see `useEffect` being used anywhere.

iandunn (approved):
This looks great, kudos!

I left a few minor suggestions, but I don't feel strongly about any.

settings/src/components/email-address.js
```diff
@@ -0,0 +1,94 @@
...
+import { store as coreDataStore, useEntityRecord } from '@wordpress/core-data';
+
+/**
+ * Render the Email setting.
+ */
+export default function EmailAddress( { userId } ) {
+	const { record, edit, save, editedRecord, hasEdits } = useEntityRecord( 'root', 'user', userId );
+
+	// useEntityRecord() doesn't expose isSaving. Huh.
+	const isSaving = useSelect( ( select ) => select( coreDataStore ).isSavingEntityRecord( 'root', 'user', userId ) );
```
iandunn: @adamziel is there a way to get this from `useEntityRecord`? I don't see one either.

settings/src/components/email-address.js
```diff
@@ -0,0 +1,94 @@
...
+	};
+
+	const handleDiscard = async () => {
+		try {
+			await edit( { pending_email: '' } );
+			await save();
+		} catch( error ) {
+			alert( error.message );
+		}
+	};
```
iandunn: Should all of these use `useCallback` to avoid re-defining them during each render?

I think that's important in some cases, but it might just be a micro-optimization here.

settings/src/script.js
```diff
@@ -4,12 +4,15 @@
 import { __ } from '@wordpress/i18n';
 import { reactDOM, StrictMode, useState } from '@wordpress/element';
 import { Icon, arrowLeft } from '@wordpress/icons';
+import { Spinner } from '@wordpress/components';
+import { store as coreDataStore, useEntityRecord } from '@wordpress/core-data';
```
iandunn, line 9:
```suggestion
import { useEntityRecord } from '@wordpress/core-data';
```

It doesn't look like `store` is used in this file.

settings/src/components/email-address.js
```diff
@@ -0,0 +1,94 @@
...
+import { Button, TextControl, Notice, Spinner } from '@wordpress/components';
+import { useState, useEffect } from '@wordpress/element';
+import { useSelect } from '@wordpress/data';
+import { store as coreDataStore, useEntityRecord } from '@wordpress/core-data';
+
+/**
+ * Render the Email setting.
+ */
+export default function EmailAddress( { userId } ) {
+	const { record, edit, save, editedRecord, hasEdits } = useEntityRecord( 'root', 'user', userId );
```
iandunn: Since this will be repeated on every screen, and it's already done in `Main`, maybe should just pass the entire result from `useEntityRecord` to `CurrentScreen` ?

settings/src/script.js
dd32: I believe I originally had `isResolving` here, but ran into an edge-case where it was `false` but `userData` was still unavailable. I believe that would've been due to the order of operations of likely retrieving the userData before the status.

I haven't verified if this is resolved by using `useEntityRecord()` instead, which handles those operations in the correct order (i assume). It also offers a `hasResolved` which could be used instead.

By using `! userData` here it does leave an endless-spinner on the page in the event of the userData being unable to be retrieved (where it would show a UI with "your email is undefined"), such as when you attempt to edit a user you can't edit.. but that's been handled in PHP now instead.

settings/src/script.js
dd32: ~Alternatively; we shouldn't pass userData at all, as elements cannot change/update when the record is updated, unless they're being re-rendered. So it's safer to always retrieve the user object using `useEntityRecord()`.~

~`userData` was kept there for the `<AccountStatus />` which had no reason to retrieve an entire user object if it was known already since it's only displaying data and not allowing edits. I should just update that to accept a userId though I think.~

Ignore this, it stemmed from a misunderstanding of state, I've since updated this to pass a `userRecord` which is the result from `useEntityRecord()` which simplifies it all.

settings/src/components/email-address.js
dd32: Agreed!

I was also wondering if there was a need here to use a centralised noticeStore instead or something.. I've seen examples of the Editor interface having a single notice spot, rather than it being handled by individual components, but I guess since the notices are shown within the components, it makes sense to keep it here for now.

settings/src/components/email-address.js
dd32: I haven't learnt `useCallback` yet :)

I think you're right, it's likely micro-optimization here, but there's no reason not to use best practices for the smallest of things.

settings/src/script.js
dd32: Yep, missed in 80b437ff76794d876b8c0b882b96005bc863995c, removed!

settings/src/components/email-address.js
dd32: This seems like a reasonable suggestion, if the result set can be passed around like that.. in my early testing that wasn't possible, but it was more likely my misunderstanding of the state mechanisms.

settings/rest-api.php
dd32: TBH I don't want it going to that URL either really.. I think it should go to `login.wordpress.org/wp-login.php?confirmaction=blahblah` instead, and we could then have that redirect to a page of our choosing afterwards.

I need to make a Core PR to move the email confirmation code there instead of us adding special handling.

settings/src/script.js
dd32: The cause is that `isResolving` is only truthful once resolution has started, not before. So before it tried to fetch the record it wasn't resolved and it wasn't resolving.

Switching to `! hasResolved` works here, as would `status === SUCCESS`, but I've gone for the former.
This also allowed removing the `<Spinner />` from the child-classes, as it's safe to assume the parent has waited for the record to exist.

settings/src/components/email-address.js
dd32: For now, I've polyfill'd it in with a wrapping function in c1465bd2580fc7656af78ed3c49078be2ce03c3f which seems to work for me.

dd32:
Thanks for the review @iandunn!

I think this can be merged as-is (Please go ahead; I've just run out of time today), and TODOs completed afterwards.

dd32:
Something I noticed, not a blocker, you can leave the page during the save process by clicking the in-page Back button. The save will continue, and if successful, the Account status page updates to reflect the change. This seems odd, but also expected and `worksforme` or at least `worksformefornow`.

settings/src/script.js
iandunn: That makes sense 👍🏻

settings/rest-api.php
iandunn: 👍🏻

settings/src/script.js
```diff
@@ -81,22 +98,24 @@ function Main() {
...
-			input field
-		</p>
-	);
+function getUserRecord( userId ) {
+	let userRecord = useEntityRecord( 'root', 'user', userId );
+
+	// Polyfill in isSaving.
+	if ( undefined === userRecord.isSaving ) {
+		userRecord.isSaving = useSelect( ( select ) => select( coreDataStore ).isSavingEntityRecord( 'root', 'user', userId ) );
+	}
```
iandunn, line 116: Nice 👍🏻

settings/src/components/email-address.js
```diff
@@ -0,0 +1,107 @@
...
+			}
+
+			setEmailError( message );
+		}
+	}, [] );
+
+	const handleDiscard = useCallback( async () => {
+		try {
+			await edit( { pending_email: '' } );
+			await save();
```
iandunn, line 39: It might be a little safer to do something like `edit( record )` (and then not call `save()`. That'd guarantee that nothing is changed, and avoid an HTTP call.

I haven't tested if that actually works, though.

iandunn: Er, nevermind, this needs to `save()` to change the pending address in the db 👍🏻

settings/src/components/email-address.js
adamziel: @iandunn @dd32 This information isn't currently returned by `useEntityRecord`. The idea was to encourage using the promise returned by `save()` to track isSaving. It won't work for all use-cases, though, so feel free to propose a PR for `useEntityRecord` and ping me for reviews.

PR #33, opened 2022-12-29
Add dynamic data/interactions to Password component

### LINKED ISSUE #27: UI: Password
Implement the password field in the custom UI, see #18 for rough mockups

iandunn:
Fixes #27

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -45,9 +45,13 @@ function render_custom_ui() : void {
 	}
 
 	wp_enqueue_style( 'wp-components' );
+	wp_enqueue_script( 'zxcvbn-async' ); // Can't be enqueued via asset.php because it doesn't have a `wp-` prefix.
 
 	$user_id    = bbp_get_displayed_user_id();
-	$json_attrs = json_encode( [ 'userId' => $user_id ] );
+	$json_attrs = json_encode( [
+		'userId'          => $user_id,
+		'userRequires2fa' => user_requires_2fa( get_userdata( $user_id ) ),
+	] );
 
 	$preload_paths = [
 		'/wp/v2/users/' . $user_id . '?context=edit',
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -41,6 +41,7 @@ function replace_core_ui_with_custom() : void {
  */
 function render_custom_ui() : void {
 	if ( ! current_user_can( 'edit_user', bbp_get_displayed_user_id() ) ) {
+		echo 'You cannot edit this user.';
 		return;
 	}
 
@@ -48,10 +49,7 @@ function render_custom_ui() : void {
 	wp_enqueue_script( 'zxcvbn-async' ); // Can't be enqueued via asset.php because it doesn't have a `wp-` prefix.
 
 	$user_id    = bbp_get_displayed_user_id();
-	$json_attrs = json_encode( [
-		'userId'          => $user_id,
-		'userRequires2fa' => user_requires_2fa( get_userdata( $user_id ) ),
-	] );
+	$json_attrs = json_encode( [ 'userId' => $user_id ] );
 
 	$preload_paths = [
 		'/wp/v2/users/' . $user_id . '?context=edit',
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -189,6 +189,8 @@ function isPasswordStrong( password, userRecord ) {
 	) );
 	blocklist = blocklist.concat( [ 'wordpress', 'wporg', 'wordpressorg' ] );
 
+	// `3` is "safely unguessable: moderate protection from offline slow-hash scenario. (guesses < 10^10)"
+	// `4` is "very unguessable: strong protection from offline slow-hash scenario. (guesses >= 10^10)"
 	const minimumScore = userRecord['2fa_required'] ? 4 : 3;
 	const strength     = zxcvbn( password, blocklist );
 
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -12,16 +12,12 @@ import { useCallback, useEffect, useState }      from '@wordpress/element';
 import { Icon, cancelCircleFilled, check, seen } from '@wordpress/icons';
 import apiFetch                                  from '@wordpress/api-fetch';
 
-import {} from '@wordpress/util'; // Enqueue via asset.php.
-const { post } = wp.ajax;
-
 /**
  * Render the Password setting.
  */
 export default function Password( { userRecord } ) {
 	const [ passwordStrong, setPasswordStrong ] = useState( false );
 	const [ saved, setSaved ]                   = useState( false );
-	const [ generating, setGenerating ]         = useState( false );
 	const [ inputType, setInputType ]           = useState( 'password' );
 
 	useEffect( () => {
@@ -39,20 +35,9 @@ export default function Password( { userRecord } ) {
 	 * to update the `passwordStrong` state. Is there a way to tell React to wait until both are done to re-render?
 	 * Or maybe the condition that renders the notice can include something like `hasResolved`?
 	 */
-	const generatePasswordController = useCallback( async () => {
-		setGenerating( true );
-		let password;
-
-		try {
-			password = generatePasswordInBrowser();
-		} catch ( exception ) {
-			console.error( exception );
-			password = await post( 'generate-password' );
-		}
-
-		userRecord.edit( { password } );
+	const generatePasswordHandler = useCallback( async () => {
+		userRecord.edit( { password: generatePassword() } );
 		setInputType( 'text' );
-		setGenerating( false );
 	}, [] );
 
 	const savePassword = useCallback( async () => {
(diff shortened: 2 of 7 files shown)


settings/settings.php
```diff
@@ -45,9 +45,13 @@ function render_custom_ui() : void {
 	}
 
 	wp_enqueue_style( 'wp-components' );
+	wp_enqueue_script( 'zxcvbn-async' ); // Can't be enqueued via asset.php because it doesn't have a `wp-` prefix.
 
 	$user_id    = bbp_get_displayed_user_id();
-	$json_attrs = json_encode( [ 'userId' => $user_id ] );
+	$json_attrs = json_encode( [
+		'userId'          => $user_id,
+		'userRequires2fa' => user_requires_2fa( get_userdata( $user_id ) ),
```
dd32: I'm thinking this should be exposed on the Rest API endpoint instead of being passed through the block attributes, but that can be changed afterwards as we complete more of the 2fa dynamics

settings/src/components/password.js
```diff
@@ -1,21 +1,72 @@
...
+
+	/**
+	 * @todo When starting from a fresh page load, theres a flash of the red "too easy" notice before it shows
+	 * the success notice. It happens between the time when `edit()` is called and the `useEffect` callback fires
+	 * to update the `passwordStrong` state. Is there a way to tell React to wait until both are done to re-render?
+	 * Or maybe the condition that renders the notice can include something like `hasResolved`?
+	 */
+	const generatePassword = useCallback( async () => {
+		setGenerating( true );
+		const password = await post( 'generate-password' );
```
dd32: Although this is how Core does it, I think we can probably skip this and use an in-browser generation instead.

That'll take away some of the complexity around time-to-generate, maybe falling back (with the worse UX) for browsers that don't support the crypto extensions.

For example: [See this Stackoverflow](https://stackoverflow.com/questions/9719570/generate-random-password-string-with-requirements-in-javascript#:~:text=UPDATE%2012/2022%20re%20%22security%22).

settings/src/components/password.js
```diff
@@ -1,21 +1,72 @@
...
+		setInputType( 'text' );
+		setGenerating( false );
+	}, [] );
+
+	/**
+	 * @todo This works the first time it's called, but subsequent calls result in a `403` error. That might be
+	 * because changing the password resets the auth cookies. The failed request is automatically repeated by
+	 * `useEntityRecord.save()`, though, and that request succeeds. The only tangible problem here is the console
+	 * error, but it'd be nice to fix that.
+	 */
```
dd32: I'm not seeing this behaviour, the `POST` completes with a `200` and returns the new user object.

What's the request/contents of the request for the 403 and subsequent 200?

dd32 (approved):

settings/settings.php
iandunn: Ah, yeah, that's a good idea 👍🏻

settings/src/components/password.js
iandunn: Also a good idea 👍🏻

iandunn: ...aaaaaand that turned into a rabbit hole 😣

Even after throwing out the answers on that SO thread that don't use `crypto.getRandomValues()`, it's still not really clear what the proper way to do it is.

JS doesn't provide a native way to uniformly get a cryptographically-secure random number _within a range_, and doing it manually seems [complicated](https://stackoverflow.com/questions/41437492/how-to-use-window-crypto-getrandomvalues-to-get-random-values-in-a-specific-rang) and [error-prone](https://dimitri.xyz/random-ints-from-random-bits/).

I searched for info from a security expert about using `getRandomValues()` to generate a password, but couldn't find any. Given how nuanced the topic is and how many different methods there are, I'm leery to rely on an SO answer.

Maybe I'm being too paranoid, though, and [mwag's answer](https://stackoverflow.com/a/43020177/450127r) is Good Enough? I've cleaned it up in 0b76f5b to make it easier to read and try out. The part I'm not sure about is if it's really a uniform distribution like the author claims, or if using the modulus creates a bias.

I might take a closer look later, but I think I've probably spent too much time on it for now, and am leaning towards sticking w/ the AJAX request.

settings/settings.php
iandunn: done in d95cb3d09e0d919c25283e6f9fa121e16a0a03ce

settings/src/components/password.js
iandunn: 🤔 it happens for me in FF and Chrome.

https://user-images.githubusercontent.com/484068/210116102-002fda15-e6ed-4ad4-bb99-eb4bceb47218.mov

iandunn: fixed in 6ac3885

settings/src/components/password.js
dd32: Using `%` does introduce a bias, but given the small array size of characters, it's not going to be _that bad_ imho.

I certainly wouldn't rely upon it for key generation, but for a psuedo-random password it seems _good enough_.

Realistically, using random character passwords is not ideal regardless, as most people are more likely to use their own password OR they use a password manager which would insert a much better random password, rather than what we should see as a **strong password**. We don't want people using the random generated passwords, and NOT using a password manager..

I took a look what Automattic uses on Calypso, and this is what I came up with: https://github.com/Automattic/wp-calypso/blob/trunk/client/lib/generate-password/index.js

Is there a 3rd-party library we can just import and rely upon for offering stronger passwords perhaps? Maybe that generates it based on a wordlist? 

Sticking with the XHR for now is fine by me, if it's easier, it's probably worthwhile looking at alternatives long-term.

settings/src/components/password.js
dd32: Ahhhh this is an issue on local environments, but not on WordPress.org, as WordPress.org doesn't have the nonces tied to the login session.

This sounds like a bug though, The `X-WP-Nonce` header returned on the `POST wp-json...` endpoint should be updated to the new value, it might be that it's used for future REST API's but not for admin-ajax endpoints.. OR the `set password` code in the rest api is re-generating the nonce prior to the new password being set, that invalidates the existing session.

settings/src/components/password.js
dd32: Looks like PHP's `random_int()` does something similar to 0b76f5bac57608aa40b22d1d99c6ee0e84d91d7f - It just keeps generating numbers until something falls within the range: https://github.com/php/php-src/blob/master/ext/random/random.c#L623-L665 although it has some optimization as powers of two aren't biased apparently.

iandunn: Yeah, I agree it's ideal if folks use a password manager or their browser to generate it, since those are in a better position to ensure randomness. In a few cases the entropy may be lower on paper, like when Firefox generates passwords that are only 15 characters, but that's probably not a big deal 🤷🏻‍♂️ 

For a pass_phrase_ library, [niceware](https://www.npmjs.com/package/niceware) seems like the best from a security perspective, but it's completely unusable for us since their dictionary contains vulgar words. Using the [EFF large wordlist](https://www.eff.org/deeplinks/2016/07/new-wordlists-random-passphrases) would be a requirement IMO. [eff-diceware-passphrase](https://www.npmjs.com/package/eff-diceware-passphrase) also looks good, but is written for Node, and porting it along with the related dependencies wouldn't be worth it. I couldn't find anything else that I'd trust, unfortunately 🙁

For a pass_word_ library, [password-generator](https://www.npmjs.com/package/password-generator) looks pretty good based on usage, but I didn't see anything using `crypto.getRandomValues()` written by someone that I can tell is an expert.

iandunn: ...oh, wait, looking at the code there, it looks like if we just use a `UInt8Array` instead of a `UInt32Array` then the random numbers are bounded to `0-255` 🤦🏻‍♂️ 

So then it's performant to use rejection sampling to avoid bias, which is what that library does. I went ahead and did that in 5241afe

settings/src/components/password.js
iandunn: > not [an issue] on WordPress.org, as WordPress.org doesn't have the nonces tied to the login session.

Ah, ok 👍🏻 

> it might be that it's used for future REST API's but not for admin-ajax endpoints

Were you thinking this was interacting w/ the `generate-password` AJAX endpoint (which was removed in https://github.com/WordPress/wporg-two-factor/commit/5241afeac49739ba58d8162f468434291dbc0fb6)? I can see those not being compatible, but that doesn't seem to be an issue here since all the requests in the `savePasswordHandler()` use `userRecord.save()`.

I'm guessing it's harmless to leave this on in production, but if you think it's best I could look into passing `wp_get_environment_type()` to JS, and wrapping this in a conditional.

dd32: Nah Leave it in production, we'll need it eventually.

My point was that in the event of `user.save()` the new Authentication cookies are served through the rest api, but, the nonces weren't being refreshed automatically for you seemingly (although the REST API responds with a new nonce as part of the `.save()` - see the response headers for the x-wp-nonce).

I haven't tested your scenario, I was spit-balling, I don't know which of the nonces are invalid (one, or both) or if it should be automatically updating it or not. My gut-feeling is that the response should be responding with a new ajax nonce on the rest api XHR potentially, but the presence of the `refreshNonce` API seems.. well I guess it does the trick.

xknown:
Out of curiosity, why are doing this on the client side? Do we have the same checks on the backend that block weak or compromised passwords?

iandunn:
Building a custom UI was one of the requirements of the project, since the design of the wp-admin UI in the upstream plugin isn't modern enough.

I did initially use the `generate-password` AJAX endpoint to generate the password, but we [decided to switch to doing it in the browser](https://github.com/WordPress/wporg-two-factor/pull/33#discussion_r1059228749) for performance/UX.

The _client_ does check [check the strength](https://github.com/WordPress/wporg-two-factor/blob/fca9533af9e43d639078a83d822701554770d021/settings/src/components/password.js#L23-L31), and [blocks saving](https://github.com/WordPress/wporg-two-factor/blob/fca9533af9e43d639078a83d822701554770d021/settings/src/components/password.js#L115-L117) when it's too weak. [It's checked with zxcvbn](https://github.com/WordPress/wporg-two-factor/blob/fca9533af9e43d639078a83d822701554770d021/settings/src/components/password.js#L166-L190), and we 
 [require a higher entropy score](https://github.com/WordPress/wporg-two-factor/blob/fca9533af9e43d639078a83d822701554770d021/settings/src/components/password.js#L184-L186) for folks with elevated privileges. zxcvbn is somewhat outdated, but it's what Core still uses. If you feel like that's not good enough,  I'd be open to swapping in a different one.

The REST endpoint _doesn't_ check the strength when saving, so somebody could intentionally set a weak password if they really wanted to. But, we do have some backend code in a separate plugin that monitors for weak/compromised passwords, and would force them to reset it.

settings/src/components/password.js
iandunn: That sounds good 👍🏻

xknown:
> The client does check [check the strength](https://github.com/WordPress/wporg-two-factor/blob/fca9533af9e43d639078a83d822701554770d021/settings/src/components/password.js#L23-L31), and [blocks saving](https://github.com/WordPress/wporg-two-factor/blob/fca9533af9e43d639078a83d822701554770d021/settings/src/components/password.js#L115-L117) when it's too weak. [It's checked with zxcvbn](https://github.com/WordPress/wporg-two-factor/blob/fca9533af9e43d639078a83d822701554770d021/settings/src/components/password.js#L166-L190), and we
[require a higher entropy score](https://github.com/WordPress/wporg-two-factor/blob/fca9533af9e43d639078a83d822701554770d021/settings/src/components/password.js#L184-L186) for folks with elevated privileges. zxcvbn is somewhat outdated, but it's what Core still uses. If you feel like that's not good enough, I'd be open to swapping in a different one.
> 
> The REST endpoint doesn't check the strength when saving, so somebody could intentionally set a weak password if they really wanted to. But, we do have some backend code in a separate plugin that monitors for weak/compromised passwords, and would force them to reset it.

While I understand the motivations of doing this only client side, I think it'd be good to also enforce this on the backend, especially if we have already existing code to check for compromised/weak passwords -- it wouldn't be a good UX to set a password via this endpoint, and then later be blocked because it was considered weak or compromised somewhere else. Also, doing this only client side will most likely be reported in our bounty program :)

iandunn:
👍🏻 , I added backend validation to the endpoint in `r19805-dotorg`.

dd32:
> > While I understand the motivations of doing this only client side, I think it'd be good to also enforce this on the backend,
>
> I added backend validation to the endpoint in r19805-dotorg.

It's worth noting that WordPress core / wp-admin doesn't do this, you can set any password as all the validation is client-side. As a result, we'll have to disable the admin UIs, I've opened #37 to track that.

PR #36, opened 2023-01-05
[WIP] Add dynamic TOTP functionality

### LINKED ISSUE #30: UI: TOTP Setup
Implement the TOTP fields in the custom UI, see #18 for rough mockups

iandunn:
Fixes #30 

This still has some rough edges, but works pretty well. We'll definitely need to iterate on the progress bar design, but I think it'll be helpful to get the functionality merged and then circle back to that.

To test:

1. disable TOTP via `wp-admin/profile.php`
2. visit `/users/{username}/edit/account/`. the status should show that it's disabled.
3. click on totp screen and follow process to setup using qr code
4. repeat steps using text code instead of qr code
5. repeat steps using keyboard instead of mouse


pkevan:
Seems like there is a dependancy on https://github.com/WordPress/two-factor/pull/504 since testing the code as-is via a sandbox produces a fatal error when viewing `/users/{username}/edit/account/`:

`Fatal error: Uncaught Error: Call to undefined method Two_Factor_Totp::generate_qr_code_url()`

iandunn:
Doh, I thought I'd mentioned that in the issue, but I didn't.

iandunn:
I've got #42 ready, and it depends on this being merged, so I'm gonna go ahead and merge it so we can keep moving. If you have any thoughts, though, please add them and we can always iterate. Reading & commenting on the code is more valuable than testing IMO, if you only have time for one of them.

Issue #52, opened 2023-02-08
Prevent changing 2FA settings through wp-admin


iandunn:
We already hide the UI that the `two-factor` plugin adds to `profile.php`, but our custom UI also includes password/email, and those are still in wp-admin/buddypress/bbpress/etc.

> there's a lot of settings on wp-admin/profile.php that should not be changed there, and in the case of WordCamp, will cause the WordPress.org user data to get out of sync with the database.
> https://github.com/WordPress/wporg-two-factor/issues/13#issuecomment-1317931757

- [ ] If a setting exists in the custom wporg-two-factor UI, then it'd ideally be removed from wp-admin/buddypress/bbpress/etc profile editors, and replaced w/ a link to the 2FA settings screen
- [ ] `wp_user*` database values on sites that share user tables (wordcamp, buddypress, others?) should also be replaced w/ a link to the w.org profile editor

We might need to delay this while #114 is still using wp-admin, before the custom UI is built.

dd32:
I've gone back and forth on this, the wp-admin profile page is not accessible by most users, but is still required for Application Passwords (As used by the PHPUnit Test Reporters).

It feels like a complete hack, but filters and blocking change attempts doesn't seem to be worth it, and I don't like the look of the code.. so.. I'm super temped to just deploy some Javascript to remove the fields we don't want changed, along with a big banner.. That's what has resulted in me making #54.

dd32:
Between #54 and b214690eb5ffa7c88db428f87980632eafa65464 I'm calling this resolved.

PR #54, opened 2023-02-09
Add a notice to wp-admin/profile.php on where to edit your profile

dd32:
See #52

I'm not super happy about this approach, but due to the lack of filters in user-edit.php to allow simply removing the fields, this seemed like the best approach.

There's a carve-out for setting the user role for the site, as that still may need to be set on the individual user pages.

The profile.php & user-edit.php pages still need to exist in some form right now, if not only for the fact that it's the only place to create Application Passwords, which while not used by many, are used by things like the PHPUnit Test Reporters.

We no longer link to `profile.php` except within the admin-bar when logged into wp-admin, so this approach feels good enough to me.


settings/settings.php
```diff
@@ -100,7 +138,7 @@ function render_custom_ui() : void {
 	echo do_blocks( "<!-- wp:wporg-two-factor/settings $json_attrs /-->" );
 }
 
-/**
+/*
```
dd32:
```suggestion
/**
```

PR #60, opened 2023-02-14
TOTP: Add space to token placeholder for readablility.

iandunn:
This updates our custom UI to match the changes in https://github.com/WordPress/two-factor/pull/518/

This replace Gutenberg's `NumberControl` with a new `NumericControl`, because the former will not allow spaces.

### DIFF
--- a/settings/src/components/numeric-control.js
+++ b/settings/src/components/numeric-control.js
@@ -19,6 +19,7 @@ export default function NumericControl( props ) {
 			inputMode="numeric"
 			autoComplete={ props.autoComplete || 'off' }
 			pattern={ props.pattern || "[0-9 ]*" }
+			title={ props.title || "Only numbers and spaces are allowed" }
 			onChange={ ( event ) => {
 				// Most callers will only need the value, so make it convenient for them.
 				props.onChange( event.target.value, event );
--- a/settings/src/components/numeric-control.js
+++ b/settings/src/components/numeric-control.js
@@ -1,0 +1,28 @@
+/**
+ * Input field for values that use digits, but aren't strictly numbers in the mathematical sense.
+ *
+ * The classic example is a credit card, but in our context we have TOTP codes, backup codes, etc. We may want to
+ * display them with spaces for easier reading, etc.
+ *
+ * If we used Gutenberg's `NumberControl`, we'd have to hide the extraneous UI elements, and it would still be
+ * using the underlying `input[type="number"]`, which has some accessibility issues.
+ *
+ * @link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#accessibility
+ * @link https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/
+ * @link https://stackoverflow.com/a/66759105/450127
+ */
+export default function NumericControl( props ) {
+	return (
+		<input
+			{ ...props }
+			type="text"
+			inputMode="numeric"
+			autoComplete={ props.autoComplete || 'off' }
+			pattern={ props.pattern || "[0-9 ]*" }
+			onChange={ ( event ) => {
+				// Most callers will only need the value, so make it convenient for them.
+				props.onChange( event.target.value, event );
+			} }
+		/>
+	);
+}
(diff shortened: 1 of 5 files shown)


settings/src/components/totp.scss
```diff
@@ -21,7 +21,7 @@
 }
 
 .wporg-2fa__verify-code {
-	width: 10ch;
+	width: 13ch;
```
StevenDufresne: Can we remove this entirely? The input feels tight. 

Additionally, it seems like bbpress is overriding CSS padding. Can we update it to `padding: 6px 10px` to match the text styles from `_fields.scss`?



| Before | After |
| --- | --- |
| Screen Shot 2566-03-10 at 13 45 36 | Screen Shot 2566-03-10 at 13 45 25  |

settings/src/components/numeric-control.js
```diff
@@ -0,0 +1,28 @@
...
+ * @link https://stackoverflow.com/a/66759105/450127
+ */
+export default function NumericControl( props ) {
+	return (
+		<input
+			{ ...props }
+			type="text"
+			inputMode="numeric"
+			autoComplete={ props.autoComplete || 'off' }
+			pattern={ props.pattern || "[0-9 ]*" }
```
StevenDufresne, line 21: Consider adding a title to explain the pattern:

`title="Verification codes only include numbers"`

Gives:
Screen Shot 2566-03-10 at 13 54 36

settings/src/components/totp.js
```diff
@@ -197,28 +198,18 @@ function createQrCode( data ) {
  */
 function SetupForm( { handleEnable, verifyCode, setVerifyCode, qrCodeUrl, secretKey } ) {
 	const verifyCodeLength = 6;
-	const canSubmit        = qrCodeUrl && secretKey && verifyCode && verifyCode.length === verifyCodeLength;
+	const cleanVerifyCode  = verifyCode.replaceAll( /\s/g, '' );
```
StevenDufresne, line 201: Should we remove everything except numbers?

StevenDufresne: I would also consider not updating the value after `verifyCodeLength` is reached. 

It's really easy to add an extra number and slip by the "enabled" state on the button and since the placeholder is gone, I would only have the label above the input to inform my next steps... but it's not likely that I read that carefully.

iandunn: > Should we remove everything except numbers?

I worry that could confuse the user; they might not understand why what they type wasn't showing up, and think there's a bug or something. Letting them see the `pattern/title` warning seems like a better UX.

> not updating the value after verifyCodeLength is reached

That seems more intuitive, but maybe has a similar issue to above. I played around with automatically triggering the warning if the length is too much, but then you have to hide it again when they backspace, and that opens questions about if it should appear when the count is < `6`... It starts to feel like we're overcomplicating it.

I'm open to it if you see an elegant way to do it, though.

iandunn: We could maybe just remove the `canSubmit` bit entirely, so that users can always click it, and when they do they'll see the `title` warning?

settings/src/components/numeric-control.js
iandunn: Great catch! I added that in fd77cf3

settings/src/components/totp.scss
iandunn: 👍🏻 added in fd77cf3

Without any `width` the input is much longer for me, so I kept it, but increased it a lot.

settings/src/components/totp.js
StevenDufresne: > > Should we remove everything except numbers?
> 
> I worry that could confuse the user; they might not understand why what they type wasn't showing up, and think there's a bug or something. Letting them see the `pattern/title` warning seems like a better UX.

👍 

> 
> > not updating the value after verifyCodeLength is reached
> 
> That seems more intuitive, but maybe has a similar issue to above. I played around with automatically triggering the warning if the length is too much, but then you have to hide it again when they backspace, and that opens questions about if it should appear when the count is < `6`... It starts to feel like we're overcomplicating it.
> 
> I'm open to it if you see an elegant way to do it, though.

Yeah, I don't think triggering the warning is necessary, it would introduce too much to take in. I guess this is why some applications have the separate inputs. Ie: `[ ] [ ] [ ]  [ ] [ ] [ ]`.

This is not a deal breaker.

iandunn: Yeah, it'll be much nicer when/if we can [just use CSS to add a space in the middle](https://github.com/WordPress/two-factor/pull/518#discussion_r1110126364).

PR #62, opened 2023-02-16
Fix basic accessibility issues

### LINKED ISSUE #61: Accessibility improvements
Issues found:

1. No focus styling on profile page Account action buttons
2. Password page 'show password' button has no accessible name
Screen Shot 2023-02-15 at 10 55 14 AM
3. No focus styling on 2FA page QR code or cancel button 
4. 2FA page QR code has no discernible text
Screen Shot 2023-02-15 at 10 50 23 AM
5. Success notifications when focused say 'Dismiss this notice, button' but they aren't dismissable via mouse or keyboard
Screen Shot 2023-02-15 at 11 17 31 AM

Questions

1. Should we add UI which is keyboard accessible for easy copy and paste of backup codes?
2. Should the 'password best practices' link open in a new tab? Feels disruptive to navigate directly to an info link.

adamwoodnz:
- [x] No focus styling on profile page Account action buttons
- [x] Password page 'show password' button has no accessible name
- [x] No focus styling on 2FA page QR code or cancel button 
- [x] 2FA page QR code has no discernible text
- [x] Success notifications when focused say 'Dismiss this notice, button' but they aren't dismissable via mouse or keyboard

### Screenshots

Focus styling on account action buttons

Screen Shot 2023-02-21 at 3 12 46 PM

Focus styling on QR code and cancel button

Screen Shot 2023-02-21 at 3 12 20 PM

Fix dismissal of snackbar notification

Screen Shot 2023-02-21 at 4 16 45 PM

NOTE: I followed the [design](https://github.com/WordPress/wporg-two-factor/issues/18#issue-1461452765) for this, which shows a dismiss button. The suggested `isDismissible` prop doesn't exist on the [Snackbar](https://developer.wordpress.org/block-editor/reference-guides/components/snackbar/), but `explicitDismiss: true` displays the dismiss button.

Screen Shot 2023-02-21 at 4 18 28 PM


settings/src/components/account-status.js
```diff
@@ -65,9 +65,16 @@ export default function AccountStatus() {
 /**
  * Render a card for the status of the given setting.
  */
-function SettingStatusCard( { screen, status, headerText, bodyText } ) {
+function SettingStatusCard({ screen, status, headerText, bodyText }) {
+	const [ isFocused, setIsFocused ] = useState( false );
+	const className = [
+		"wporg-2fa__status-card",
+		"wporg-2fa__status-card-" + screen,
+		...( isFocused ? [ "wporg-2fa__status-card__focused" ] : [] ),
```
adamwoodnz: This classname is used to apply `z-index` to the card (see [styles](https://github.com/WordPress/wporg-two-factor/pull/62/commits/1a46841a12fbae38f68af39d956561d8240d4ac5#diff-41c16ca64f385109d400c93a6d9fe13d0b8d2cab74a3381a361215a61ba72b94)), lifting it to the top of the stack and allowing the focus outline to be shown. The focus event fires on the nested anchor element so a state setter is passed down to be called on focus and blur.

Without the `z-index` it looks like this

Screen Shot 2023-02-21 at 3 05 23 PM

settings/src/components/screen-link.js
```diff
@@ -28,6 +34,9 @@ export default function ScreenLink( { screen, anchorText, buttonStyle = false }
 			href={ screenUrl.href }
 			onClick={ ( event ) => clickScreenLink( event, screen ) }
 			className={ classes.join( ' ' ) }
+			tabIndex={ 0 }
```
adamwoodnz: The `onBlur` event won't fire without this

settings/src/components/screen-link.scss
```diff
@@ -1,6 +1,10 @@
 /* Overwrite bbPress styles with Gutenberg styles */
-#bbpress-forums .wp-block-wporg-two-factor-settings a.components-button,
-#bbpress-forums .wp-block-wporg-two-factor-settings a.components-button:hover {
+#bbpress-forums .wp-block-wporg-two-factor-settings a.components-button.is-secondary {
 	box-shadow: inset 0 0 0 1px var(--wp-components-color-accent,var(--wp-admin-theme-color,#007cba));
 	text-decoration: none;
+
+	&:not(:disabled):focus {
+		box-shadow: inset 0 0 0 var(--wp-admin-border-width-focus) var(--wp-components-color-accent,var(--wp-admin-theme-color,#007cba));
+		outline: 3px solid transparent;
```
adamwoodnz, line 8: Matches Gutenberg button style

Screen Shot 2023-02-21 at 4 26 10 PM

settings/src/components/account-status.js
iandunn: Would [`:focus-within`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within) let us avoid the need to manually track the focus?

If not, another option might be moving the `<Card>` inside `ScreenLink.anchorText`.

iandunn (approved):
This looks great, thanks!

If it turns out to be possible to remove the manual focus state tracking that'd be great, but it's not a blocker.

adamwoodnz:
> If it turns out to be possible to remove the manual focus state tracking that'd be great

Nice suggestions on that, I'll take a look 👍

settings/src/components/account-status.js
adamwoodnz: > Would [`:focus-within`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within) let us avoid the need to manually track the focus?

Yep, works! TIL, thanks

iandunn: 🎉 it looks much simpler now 🙂

PR #65, opened 2023-02-22
Revert to script with dependencies to fix broken global header nav

adamwoodnz:
The temporary fix we added to fix dependency enqueuing for this plugin, has stopped dependencies of the global header being enqueued.

### DIFF
--- a/settings/src/block.json
+++ b/settings/src/block.json
@@ -17,7 +17,7 @@
 	"textdomain": "wporg",
 	"editorScript": "file:./index.js",
 	"editorStyle": "file:./index.css",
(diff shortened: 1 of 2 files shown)


dd32:
I've committed a workaround in #66 instead, but let's leave this open.

If there's ANY further issues at all, just merge this and we can re-open the meta ticket.

dd32 (approved):
Just approving this, as a sign of looks good to me if there's any later issues with scrips.

iandunn:
I'm gonna go ahead and close this since we haven't had any problems with the workaround in #66. Feel free to reopen if you see a need though.

Issue #53, opened 2023-02-09
Handle errors in `savePasswordHandler` response


iandunn:
It's possible for a new password to pass the client-side checks, but fail the server-side ones. Currently when that happens the user doesn't get any feedback that the XHR failed. Instead we should show an error letting them know the pw is too weak.

It'd be nice to check it on the fly as they type, so that the server-side and client-side checks could be combined. That could result in too many requests to the external service, though. We could debounce it, but this error probably won't happen often, so it's probably good enough that we just show the user the error message.

PR #71, opened 2023-02-23
Convert password reset parent container into a `<form>`.

StevenDufresne:
This PR gives users the ability to update their password via pressing the enter key.

## Changes
1. Turn `passwordStrong` to be a state variable
2. Move form handling from `<button>` to `<form>` event handler. 

## Unit tests
I've added some unit tests to protect these code changes. Ideally, the `<form>` component would be its own component which would make testing easier but I didn't want to make those changes here. As a result, I couldn't easily test everything, I couldn't figure out how to test the `onSubmit` change here using `mocks` and `spies`. 🤔 

It also will help making changes in #53.

## How to test
**Shouldn't submit when field is `empty || has a weak password`**
1. Tab/Click into input field
2. Hit enter
4. Form should not submit
5. Type password "pass"
6. Hit 'enter'
7. Form should not submit

**Should submit - Enter**
1. Tab/Click into input field
2. Type a secure password
3. Hit 'enter'
4. Form should submit

**Should submit - Button Click**
1. Tab/Click into input field
2. Type a secure password
3. Click 'enter'
4. Form should submit

**Should submit - Generate passsword**
1. Click "Generate strong password"
2. Slick "Save password"
3. Form should submit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,10 @@ WordPress.org-specific customizations for the Two Factor plugin
 1. Open `wp-admin/options-general.php?page=bbpress` and uncheck `Prefix all forum content with the Forum Root slug (Recommended)`, then save.
 1. Visit https://example.org/users/{username}/edit/account/ to view the custom settings UI. If you get a `404` error, visit `wp-admin/options-permalinks.php` and then try again.
 
+## Testing
+
+Front-end unit tests can be run in `/settings` using the `npm run test:unit` command.
+
 ## Security
 
 Please privately report any potential security issues to the [WordPress HackerOne](https://hackerone.com/wordpress) program.
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ WordPress.org-specific customizations for the Two Factor plugin
 
 Front-end unit tests can be run in `/settings` using the `npm run test:unit` command.
 
+Back-end unit tests can be run in `/` using the `composer run test` or `composer run test:watch` commands. `composer run coverage` will generate a coverage report.
+
 ## Security
 
 Please privately report any potential security issues to the [WordPress HackerOne](https://hackerone.com/wordpress) program.
--- a/settings/package.json
+++ b/settings/package.json
@@ -15,10 +15,10 @@
 		"test:unit": "wp-scripts test-unit-js"
 	},
 	"dependencies": {
-		"@automattic/generate-password": "^0.1.0",
-		"@testing-library/react": "^14.0.0"
+		"@automattic/generate-password": "^0.1.0"
 	},
 	"devDependencies": {
+		"@testing-library/react": "^14.0.0",
 		"@wordpress/api-fetch": "^6.24.0",
 		"@wordpress/base-styles": "^4.17.0",
 		"@wordpress/components": "^23.4.0",
--- a/settings/package.json
+++ b/settings/package.json
@@ -11,14 +11,21 @@
 		"lint:css": "wp-scripts lint-style",
 		"lint:js": "wp-scripts lint-js",
 		"packages-update": "wp-scripts packages-update",
-		"start": "wp-scripts start --webpack-copy-php"
+		"start": "wp-scripts start --webpack-copy-php",
+		"test:unit": "wp-scripts test-unit-js"
 	},
 	"dependencies": {
-		"@automattic/generate-password": "^0.1.0"
+		"@automattic/generate-password": "^0.1.0",
+		"@testing-library/react": "^14.0.0"
 	},
 	"devDependencies": {
+		"@wordpress/api-fetch": "^6.24.0",
 		"@wordpress/base-styles": "^4.17.0",
-		"@wordpress/icons": "^9.14.0",
+		"@wordpress/components": "^23.4.0",
+		"@wordpress/core-data": "^6.4.0",
+		"@wordpress/data": "^8.4.0",
+		"@wordpress/element": "^5.4.0",
+		"@wordpress/icons": "^9.18.0",
 		"@wordpress/scripts": "^24.6.0"
 	}
 }
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -27,11 +27,7 @@ import { GlobalContext } from '../script';
 export default function Password() {
 	const { setGlobalNotice, userRecord } = useContext( GlobalContext );
 	const [ inputType, setInputType ]     = useState( 'password' );
-	let passwordStrong                    = true; // Saved passwords have already passed the test.
-
-	if ( userRecord.hasEdits ) {
-		passwordStrong = isPasswordStrong( userRecord.editedRecord.password, userRecord.record );
-	}
+	const [ passwordStrong, setPasswordStrong ] = useState( true );
 
 	// Clear the "saved password" notice when password is being changed.
 	useEffect( () => {
@@ -41,6 +37,10 @@ export default function Password() {
 
 		setGlobalNotice( '' );
 	}, [ userRecord.hasEdits ] );
+
+	useEffect( () => {
+		setPasswordStrong( isPasswordStrong( userRecord.editedRecord.password, userRecord.record ) );
+	}, [ userRecord.editedRecord.password ] );
 
 	/**
 	 * Handle clicking the `Generate Password` button.
@@ -55,9 +55,14 @@ export default function Password() {
 		setInputType( 'text' );
 	}, [] );
 
-	// Handle clicking the `Save Password` button.
+	// Handle form submission.
 	const handleFormSubmit = useCallback( async ( event ) => {
 		event.preventDefault();
+
+		if ( ! passwordStrong || userRecord.isSaving ) {
+			return;
+		}
+
 		await userRecord.save();
 
 		// Changing the password resets the nonce, which causes subsequent API requests to fail. `apiFetch()` will
@@ -69,7 +74,7 @@ export default function Password() {
 		apiFetch.nonceMiddleware.nonce = await response.text();
 
 		setGlobalNotice( 'New password saved.' );
-	}, [] );
+	}, [ passwordStrong ] );
 
 	return (
 		<form onSubmit={ handleFormSubmit }>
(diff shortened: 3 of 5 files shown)


iandunn:
Huh, that's odd, i know i tested the `Should submit - Enter` scenario, and had to make some changes to make it work. I wonder . I just tested again though and I can verify that it's not working :(

iandunn:
It may have been 0e9017eef0a7c5d24156e11d7c1771fcdcf8849d that broke the `Should submit - Enter` scenario

settings/src/components/tests/password/mocks.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {number} returnValue The score that is returned by the mocked zxcvbn library.
+ * @return {Object}
+ */
+export const mockPasswordEstimator = ( returnValue = 0 ) => {
+	// Mock the zxcvbn library
+	global.window = Object.create( window );
+	Object.defineProperty( window, 'zxcvbn', {
+		value: () => {
+			return {
+				score: returnValue, // at less than 4
```
iandunn:
```suggestion
				score: returnValue, // at less than or equal to 4
```

[The valid range](https://github.com/dropbox/zxcvbn#usage) is `0-4` inclusive

settings/src/components/tests/password/password.test.js
```diff
@@ -0,0 +1,177 @@
...
+		it( 'should display the strong password notice', () => {
+			// State: the user has updated their password to something strong
+			// although the strength is not tested here.
+			mockContext.userRecord.editedRecord.password = '@#4asdf34asdfasdf';
+			mockContext.userRecord.hasEdits = true;
+
+			useContext.mockReturnValue( mockContext );
+
+			// We'll mock the password estimator to return a score of 10.
+			mockPasswordEstimator( 10 );
```
iandunn:
```suggestion
			// We'll mock the password estimator to return a score of 4.
			mockPasswordEstimator( 4 );
```

That's the largest possible value zxcvbn will actually return

settings/package.json
```diff
@@ -11,14 +11,21 @@
 		"lint:css": "wp-scripts lint-style",
 		"lint:js": "wp-scripts lint-js",
 		"packages-update": "wp-scripts packages-update",
-		"start": "wp-scripts start --webpack-copy-php"
+		"start": "wp-scripts start --webpack-copy-php",
+		"test:unit": "wp-scripts test-unit-js"
```
iandunn, line 15: Do you mind adding a section to the repo's `README`  that mentions this command and the PHP equivalent?

settings/src/components/tests/password/password.test.js
```diff
@@ -0,0 +1,177 @@
...
+	it( 'should submit form on button press', () => {
+		// State: the user has updated their password to something strong
+		// although the strength is not tested here.
+		mockContext.userRecord.editedRecord.password = '@#4asdf34asdfasdf';
+		mockContext.userRecord.hasEdits = true;
+
+		useContext.mockReturnValue( mockContext );
+
+		// We'll mock the password estimator to return a score of 10.
+		mockPasswordEstimator( 10 );
```
iandunn:
```suggestion
		// We'll mock the password estimator to return a score of 4.
		mockPasswordEstimator( 4 );
```

settings/src/components/tests/password/password.test.js
```diff
@@ -0,0 +1,177 @@
...
+					{ children }
+				</GlobalContext.Provider>
+			),
+		} );
+
+		const buttons = getAllByRole( 'button' );
+		const saveButton = buttons.filter(
+			( button ) => button.type === 'submit'
+		)[ 0 ];
+		fireEvent.click( saveButton );
```
iandunn, line 176: Is there a way to also test the `enter` key?

settings/src/components/password.js
```diff
@@ -27,11 +27,7 @@ import { GlobalContext } from '../script';
 export default function Password() {
 	const { setGlobalNotice, userRecord } = useContext( GlobalContext );
 	const [ inputType, setInputType ]     = useState( 'password' );
-	let passwordStrong                    = true; // Saved passwords have already passed the test.
-
-	if ( userRecord.hasEdits ) {
-		passwordStrong = isPasswordStrong( userRecord.editedRecord.password, userRecord.record );
-	}
+	const [ passwordStrong, setPasswordStrong ] = useState( true );
```
iandunn: I'm not sure that this and the related changes should be necessary, see 0e9017eef0a7c5d24156e11d7c1771fcdcf8849d. I'm curious to hear your thoughts though 'cause I could easily be wrong.

iandunn (commented):
Nice job setting up automated tests!

I left some minor comments and those aren't blockers. The only potentially significant thing is re-adding the state, but I don't have a strong opinion.

settings/src/components/tests/password/password.test.js
StevenDufresne: Usually there is, but as I mention in the description, I couldn't get it to work :). Which is ironic because it's what it want to test. 

I would be able to do it using [enzyme](https://enzymejs.github.io/enzyme/) but Gutenberg recently removed it from their project and while I didn't look into why they chose to, I would assume we should follow suit. It's better to not mount/shallow render components anyways. 

With that being said, here are things that didn't work.

- Direct: Spy on the prototype, ie `spyOn( Password.prototype, 'formSubmitHandler' )`
  - Unfortunately a functional component doesn't have a prototype
- Indirect: Tried to track calls of `userRecord.save` since it's in the `formSubmitHandler` function.
  - While `formSubmitHandler` would get called this mock would always be empty
    - I assumed it was due to the async call not returning a promise, but even after attempting that it still would show no calls. I admittedly gave up.

Ideally, the form is a component and we pass in the onSubmit handler.  We could just pass a mock `<Form onFormSubmit={ mockCallback } />` and test the call, but I didn't want to restructure this component in this pass.

StevenDufresne: What is bizarre though is that clicking the save button does register calls on the save function. 🤔 Maybe I'll take a look with fresh eyes.

settings/src/components/password.js
StevenDufresne: Yeah, that's true, as it sits, we don't technically need to add a state variable for password strength since the provider is going re-render. I ended up making these changes while looking into #53. 

Since the API could kick back a weak password error, we'll want to reshow the weak password notice. To do so, I'm thinking we'll need to wrap the `await userRecord.save()` in a `try/catch` and retrigger the weak password notice. The way this is currently working, we won't be able to do so since it only react to changes in the provider.  

**Example**
```
const savePasswordHandler = useCallback(async () => {
  try {
    await userRecord.save();

    const response = await apiFetch({
      url: apiFetch.nonceEndpoint,
      parse: false,
    });
    apiFetch.nonceMiddleware.nonce = await response.text();

    setGlobalNotice("New password saved.");
  } catch (e) {
    // check for weak password server error
    if (e.code === "password_too_weak") {
        setPasswordStrong( false )
    }

  }
}, []);
```
Does that make sense? Is there a different approach that you think may work?

settings/src/components/tests/password/password.test.js
StevenDufresne: It turns out co-pilot was confidently suggesting `keyDown` with the enter key when `submit` was the correct answer. 🧠

iandunn: Ah, nice find!

I've experienced that w/ Copilot too. Just like w/ ChatGPT, I have to spend time verifying that the solution it gives is actually correct. It still saves some time in some cases though.

settings/src/components/password.js
iandunn: Ah, that makes sense 👍🏻 Ideally those changes would be in a separate PR, but it's not a big deal if it's easier to do here. It'd be helpful for the final commit message to document that, though.

When we catch the error, will we need to create a state for the error message, in order to display it to the user? If so, then we may not need one for `passwordStrong`, since we could do something like:

```js
if ( userRecord.hasEdits ) {
	passwordStrong = isPasswordStrong( userRecord.editedRecord.password, userRecord.record );
} else if ( 'password_too_weak' === passwordSaveError.code ) {
	passwordString = false;
}
```

settings/package.json
StevenDufresne: I added an entry to the readme for front-end testing. 

I'm not certain what to add for a PHP equivalent since I only added js tests here.

settings/src/components/password.js
StevenDufresne: Yep, fair enough. I've pulled the changes from this PR. We can move the discussion over to the next one. 👍

README.md
```diff
@@ -55,6 +55,10 @@ WordPress.org-specific customizations for the Two Factor plugin
 1. Open `wp-admin/options-general.php?page=bbpress` and uncheck `Prefix all forum content with the Forum Root slug (Recommended)`, then save.
 1. Visit https://example.org/users/{username}/edit/account/ to view the custom settings UI. If you get a `404` error, visit `wp-admin/options-permalinks.php` and then try again.
 
+## Testing
+
+Front-end unit tests can be run in `/settings` using the `npm run test:unit` command.
+
```
iandunn, line 61:
```suggestion

Back-end unit tests can be run in `/` using the `composer run test` or `composer run test:watch` commands. `composer run coverage` will generate a coverage report.

```

settings/package.json
iandunn: 👍🏻 

I added a suggestion for the PHP commands.

settings/src/components/password.js
```diff
@@ -68,10 +74,10 @@ export default function Password() {
 		apiFetch.nonceMiddleware.nonce = await response.text();
 
 		setGlobalNotice( 'New password saved.' );
-	}, [] );
+	}, [ passwordStrong ] );
```
iandunn: Should this include `userRecord` too?

iandunn (approved):
This LGTM 👍🏻

settings/src/components/password.js
StevenDufresne: Yep!

PR #75, opened 2023-03-01
Turn off 2FA if the user only has Backup Codes enabled

### LINKED ISSUE #47: Disable backup codes UI until primary provider enabled
Update: The goal of this has shifted, see https://github.com/WordPress/wporg-two-factor/issues/47#issuecomment-1412483300

----

Right now we only allow regenerating backup codes, but not disabling them. The `Disable Two Factor...` button on the TOTP screen only disables TOTP. Should it also disable backup codes, or should the user be able to control them independently?

Related #21

dd32:
This is partially related to #47 and might fix it, I'm not sure.

This simply disables 2FA if the user only has Backup Codes enabled, which appears to be happening for users when _viewing the TOTP screen, but not setting it up_ (This is a bug in our UI, partially due to a less-than-ideal upstream flow).

The result from that is that they're locked out, as they never saw the backup codes, and are prompted for them next time they login. There are currently ~5 users affected by this.

This PR should resolve the login for those users, and might resolve #47.

### DIFF
--- a/wporg-two-factor.php
+++ b/wporg-two-factor.php
@@ -61,6 +61,11 @@ function set_primary_provider_for_user( string $provider, int $user_id ) : strin
 		$provider = 'Two_Factor_WebAuthn';
 	} elseif ( isset( $available_providers['Two_Factor_Totp'] ) ) {
 		$provider = 'Two_Factor_Totp';
+	} elseif ( 'Two_Factor_Backup_Codes' === $provider && 1 === count( $available_providers ) ) {
+		/*
+		 * If we only have Backup Codes enabled, 2FA is disabled on WordPress.org.
+		 */
+		$provider = '';
 	}
 
 	return $provider;
--- a/wporg-two-factor.php
+++ b/wporg-two-factor.php
@@ -34,7 +34,7 @@ add_filter( 'two_factor_totp_issuer', __NAMESPACE__ . '\set_totp_issuer' );
 add_action( 'set_current_user', __NAMESPACE__ . '\remove_super_admins_until_2fa_enabled', 1 ); // Must run _before_ all other plugins.
 add_action( 'login_redirect', __NAMESPACE__ . '\redirect_to_2fa_settings', 105, 3 ); // After `wporg_remember_where_user_came_from_redirect()`, before `WP_WPorg_SSO::redirect_to_policy_update()`.
 add_action( 'user_has_cap', __NAMESPACE__ . '\remove_capabilities_until_2fa_enabled', 99, 4 ); // Must run _after_ all other plugins.
-add_filter( 'two_factor_primary_provider_for_user', __NAMESPACE__ . '\disable_only_backup_codes', 10, 2 );
+
 
 /**
  * Determine which providers should be available to users.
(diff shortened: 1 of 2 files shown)


wporg-two-factor.php
```diff
@@ -34,7 +34,7 @@ function is_2fa_beta_tester() : bool {
 add_action( 'set_current_user', __NAMESPACE__ . '\remove_super_admins_until_2fa_enabled', 1 ); // Must run _before_ all other plugins.
 add_action( 'login_redirect', __NAMESPACE__ . '\redirect_to_2fa_settings', 105, 3 ); // After `wporg_remember_where_user_came_from_redirect()`, before `WP_WPorg_SSO::redirect_to_policy_update()`.
 add_action( 'user_has_cap', __NAMESPACE__ . '\remove_capabilities_until_2fa_enabled', 99, 4 ); // Must run _after_ all other plugins.
-
+add_filter( 'two_factor_primary_provider_for_user', __NAMESPACE__ . '\disable_only_backup_codes', 10, 2 );
```
iandunn: IMO, `two_factor_enabled_providers_for_user` might be a more intuitive/self-documenting filter to use, but this works too.

There's another `two_factor_primary_provider_for_user` callback above, so if we keep this then I think it'd be good to group the two together.

wporg-two-factor.php
```diff
@@ -239,6 +239,27 @@ function get_edit_account_url() : string {
...
 
+
+/**
+ * Disable 2FA for users who only have backup codes enabled.
+ *
+ * @param string|null $primary_provider The currently selected Primary provider.
+ * @param int         $user_id          The user ID.
+ * @return string|null The primary provider, or null.
+ */
+function disable_only_backup_codes( $primary_provider, $user_id ) {
```
iandunn: Similar to above, it'd be nice to move this next to `set_primary_provider_for_user()`, since they have the same hook.

iandunn (approved):
This works in my tests 👍🏻 

I don't think it fixes #47, since IMO we still need [the UI warning](https://github.com/WordPress/wporg-two-factor/issues/47#issuecomment-1412483300), instead of allowing the backup codes to be generated in the first place.

This is a good server-side protection to enforce that, though, and fixes it retroactively for the folks who've already experienced it.

wporg-two-factor.php
```diff
@@ -239,6 +239,27 @@ function get_edit_account_url() : string {
...
+	if ( 'Two_Factor_Backup_Codes' === $primary_provider ) {
+		$user      = get_user_by( 'id', $user_id );
+		$providers = $user ? Two_Factor_Core::get_enabled_providers_for_user( $user ) : [];
+
+		if ( 1 === count( $providers ) ) {
+			$primary_provider = null;
+		}
+	}
```
iandunn: It'd be good to add tests for the above logic.

iandunn:
I just remembered we also have #24, which is similar. cc @pkevan

pkevan:
I think I prefer this method, and since we can potentially reduce users' caps should their account fall into what we consider needing it, then it's probably a good solution.

Ideally we'd be able to guide users' easier through our preferred setup.

dd32:
I've merged it with `set_primary_provider_for_user()` as I totally missed that even existed.

I don't have an environment I can run the tests for this repo in, and the steps to get to that point are far too involved to be worth it, so I've just updated the tests with what I believe should pass.

iandunn:
> the steps to get to that point are far too involved

If you already have a Core environment locally setup, and the tests working there, then IIRC this is all that's needed:

1. Set the `WP_TESTS_DIR` env var to Core's `/tests/phpunit` folder
2. `composer install` in this plugin

I could be forgetting something, though. I'm open to changes that make it easier to setup.

iandunn:
The test was failing for me, but I fixed that in 1712d3bf05d4f6a571266835cb85e0528ceeff8d. I also pushed 0aaad94 which is unrelated but seems like a good change to make.

PR #76, opened 2023-03-06
Add a local environment for testing purposes

dd32:
This is an initial start at making this self-contained so that test can be run locally without depending on the host system having WP/phpunit configured.

To install:
`npm install`

To run (This is done as part of the install process)?
`npm env start`

To run tests:
`npm test`

See #75

PR #77, opened 2023-03-10
Update readme.

StevenDufresne:
I had to update my permalinks. Opened this as a PR to make sure it wasn't specific to my install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ WordPress.org-specific customizations for the Two Factor plugin
 1. Activate all four plugins.
 1. If you want to make JS changes, then `cd settings && npm start`
 1. Open `wp-admin/options-general.php?page=bbpress` and uncheck `Prefix all forum content with the Forum Root slug (Recommended)`, then save.
-1. Visit https://example.org/users/{username}/edit/account/ to view the custom settings UI. If you get a `404` error, visit `wp-admin/options-permalinks.php` and then try again.
+1. Visit https://example.org/users/{username}/edit/account/ to view the custom settings UI. If you get a `404` error, visit `wp-admin/options-permalinks.php`, update Permalinks to `Post name`, save and then try again.
 
 ## Testing
 


README.md
```diff
@@ -53,7 +53,7 @@ WordPress.org-specific customizations for the Two Factor plugin
 1. Activate all four plugins.
 1. If you want to make JS changes, then `cd settings && npm start`
 1. Open `wp-admin/options-general.php?page=bbpress` and uncheck `Prefix all forum content with the Forum Root slug (Recommended)`, then save.
-1. Visit https://example.org/users/{username}/edit/account/ to view the custom settings UI. If you get a `404` error, visit `wp-admin/options-permalinks.php` and then try again.
+1. Visit https://example.org/users/{username}/edit/account/ to view the custom settings UI. If you get a `404` error, visit `wp-admin/options-permalinks.php`, update Permalinks to `Post name`, save and then try again.
```
iandunn, line 56: Mine is set at `Day and name`, but this should also work. What was your set to by default? Unless it was `Plain`, then just visiting the page may have been what fixed it, since [that flushes the rewrite rules even if you don't save](https://github.com/WordPress/WordPress/blob/1bf93a87a449a9476d1efd93a0452d7d6ceb3808/wp-admin/options-permalink.php#L212).

If that's what fixed it, then we could update the readme to better explain that.

iandunn (approved):

README.md
StevenDufresne: Ah, I stood the environment up with #76. As [you mention](https://github.com/WordPress/wporg-two-factor/pull/76/files#r1131468391), this step should be added there instead. 

Just visiting doesn't work in that context, saving the permalinks doesn't either. They need to be changed?
https://d.pr/i/mpAAkI

iandunn: It looks like they're originally set to `Plain` in that gif, but I don't think rewrite rules work at all with plain permalinks. So it makes sense that they'd need to be saved in that situation.

IMO the env should setup "pretty" permalinks when it's created. cc @dd32

StevenDufresne: Yeah true.

PR #82, opened 2023-03-17
Fix notice UI for 2fa, password, and backup codes

### LINKED ISSUE #80: Improve Invalid verification code UI
The error UI can use some improvement. At the very least just adjust it's padding and make it left align with the content.

## Screenshot

Screen Shot 2023-03-10 at 2 24 17 PM

outdoor2kode:
Fixes #80 

### Password
Screen Shot 2023-03-17 at 8 30 09 AM
 
Screen Shot 2023-03-17 at 8 30 16 AM

### Two-Factor Authentication

Screen Shot 2023-03-17 at 8 29 59 AM

### Two-Factor Backup Codes

Screen Shot 2023-03-17 at 8 30 29 AM

### DIFF
--- a/settings/src/components/totp.js
+++ b/settings/src/components/totp.js
@@ -213,7 +213,7 @@ function SetupForm( { handleEnable, verifyCode, setVerifyCode, qrCodeUrl, secret
 				required={ true }
 			/>
 
-			<div>
+			<div class="wporg-2fa__verify-code wporg-2fa__button-wrapper">
 				<Button type="submit" isPrimary disabled={ ! canSubmit }>
 					Enable
 				</Button>
--- a/settings/src/components/totp.js
+++ b/settings/src/components/totp.js
@@ -213,7 +213,7 @@ function SetupForm( { handleEnable, verifyCode, setVerifyCode, qrCodeUrl, secret
 				required={ true }
 			/>
 
-			<div class="wporg-2fa__verify-code wporg-2fa__button-wrapper">
+			<div class="wporg-2fa__submit-btn-wrapper">
 				<Button type="submit" isPrimary disabled={ ! canSubmit }>
 					Enable
 				</Button>
--- a/settings/src/components/backup-codes.scss
+++ b/settings/src/components/backup-codes.scss
@@ -20,7 +20,7 @@
 	}
 
 	.components-notice {
-		margin-bottom: 1em;
+		margin: 0 0 1em;
 	}
 }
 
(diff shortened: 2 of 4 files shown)


settings/src/components/totp.js
```diff
@@ -213,7 +213,7 @@ function SetupForm( { handleEnable, verifyCode, setVerifyCode, qrCodeUrl, secret
 				required={ true }
 			/>
 
-			<div>
+			<div class="wporg-2fa__verify-code wporg-2fa__button-wrapper">
```
iandunn: Is the `width` from `wporg-2fa__verify-code` necessary on the `div`? I'd assume that the label text inside the button would set its width automatically. Hardcoding it could be a problem in the future if we start to translate the text.

Is there a semantic connection between the name/purpose of that class and this button? They don't seem to be related to me, but I could be missing it.

Since we're overriding the `margin-bottom` anyway, I wonder if it'd be better to  remove `wporg-2fa__verify-code` from this element? We could add a `width` to `wporg-2fa__button-wrapper` if there's a need for it.

iandunn (approved):
This looks good to me 👍🏻 I left a minor question/suggestion, but it's not a blocker, and this could be merged as-is.

settings/src/components/totp.js
outdoor2kode: The `wporg-2fa__verify-code` should be removed. Originally, I was thinking to put the `wporg-2fa__button-wrapper` nested inside the `verify-code` to make it more semantically connected to the `verify-code` as it's just named `button-wrapper`, which is a bit too general. 

I should have written like this
```
.wporg-2fa__verify-code {
	width: 23ch;
	margin-bottom: 20px;

	& + div {
		margin-bottom: 15px;
	}
}
```
Or give it a more precise class name so that things are making more sense. Sorry for confusing you.
And the `width` from `verify-code` is definitely not necessary, it should not be a fixed value.

I've removed `verify-code` and renamed the `button-wrapper`. If there is a better naming for the class, please let me know, I'm really bad at this.

iandunn: No worries at all, [we're all bad at it](https://martinfowler.com/bliki/TwoHardThings.html)!

iandunn (approved):
This looks even better with the latest changes 👍🏻

settings/src/components/totp.js
outdoor2kode: Classic 😆

PR #84, opened 2023-03-17
Fix "show password" icon styles

### LINKED ISSUE #83: Adjust styles of the "Show password" icon
Ref: https://github.com/WordPress/wporg-two-factor/issues/5#issuecomment-1415629776

- [x] The "Show password" icon is better to be inside the input field.
image
- [x] Have a toggled state for this icon.
image
- [x] The focus style here is a bit awkward. (It could be square if the icon sat inside the input)
image
- [x] The input should not be pushed when toggling state.
Screen Capture on 2023-03-17 at 09-00-48

outdoor2kode:
Fixes #83 

This PR fixes 
* When toggled, the "show password" icon would squeeze the input on the left.
* Make the icon inside the input field. 
* The styles were also modified to accommodate the 1password icon.
* Add a toggled state for the show password icon.
  * The `unseen` icon provided in the issue has already been added to [Gutenberg](https://github.com/WordPress/gutenberg/pull/49254), and it's expected to be in production on GB 15.5, probably 2-3 weeks later. I've opened another [branch](https://github.com/WordPress/wporg-two-factor/compare/testing/show-password-icon-style?expand=1) that had `unseen.svg` included so that it can be tested if anyone wants to.
* Run `lint:js`. 
  * It looks like there were many changes in `settings/src/components/password.js`, but most of them were just linting updates.

## Screencasts
https://user-images.githubusercontent.com/18050944/226954384-86416da1-2925-4f3f-9c2e-98ad3421a4fb.mp4

### DIFF
--- a/settings/package.json
+++ b/settings/package.json
@@ -15,8 +15,7 @@
 		"test:unit": "wp-scripts test-unit-js"
 	},
 	"dependencies": {
-		"@automattic/generate-password": "^0.1.0",
-		"lodash": "^4.17.21"
+		"@automattic/generate-password": "^0.1.0"
 	},
 	"devDependencies": {
 		"@testing-library/react": "^14.0.0",
--- a/settings/package.json
+++ b/settings/package.json
@@ -26,6 +26,7 @@
 		"@wordpress/data": "^8.4.0",
 		"@wordpress/element": "^5.4.0",
 		"@wordpress/icons": "^9.18.0",
-		"@wordpress/scripts": "^24.6.0"
+		"@wordpress/scripts": "^24.6.0",
+		"lodash": "^4.17.21"
 	}
 }
--- a/settings/package.json
+++ b/settings/package.json
@@ -15,7 +15,8 @@
 		"test:unit": "wp-scripts test-unit-js"
 	},
 	"dependencies": {
-		"@automattic/generate-password": "^0.1.0"
+		"@automattic/generate-password": "^0.1.0",
+		"lodash": "^4.17.21"
 	},
 	"devDependencies": {
 		"@testing-library/react": "^14.0.0",
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -3,7 +3,7 @@
  *
  * window.zxcvbn is also used, but can't be enqueued via an `import` because the handle doesn't have a 'wp-'
  * prefix. Instead, it's enqueued via `block.json`. It also can't be declared/destructured here because it's
- * loaded asyncronously.
+ * loaded asynchronously.
  */
 import { pick } from 'lodash';
 import { generatePassword } from '@automattic/generate-password';
@@ -200,7 +200,7 @@ export default function Password() {
  * @return {boolean} true if the password is strong, false otherwise
  */
 function isPasswordStrong( password, userData ) {
-	const { zxcvbn } = window; // Done here because it's loaded asyncronously.
+	const { zxcvbn } = window; // Done here because it's loaded asynchronously.
 
 	if ( ! zxcvbn ) {
 		return false; // Not loaded yet.
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -64,7 +64,7 @@ export default function Password() {
 	 * to update the `passwordStrong` state. Is there a way to tell React to wait until both are done to re-render?
 	 * Or maybe the condition that renders the notice can include something like `hasResolved`?
 	 */
-	const generatePasswordHandler = useCallback( async () => {
+	const handlePasswordGenerate = useCallback( async () => {
 		userRecord.edit( { password: generatePassword( 24, true, true ) } );
 		setInputType( 'text' );
 	}, [] );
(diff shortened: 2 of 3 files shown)


settings/package.json
```diff
@@ -15,7 +15,8 @@
 		"test:unit": "wp-scripts test-unit-js"
 	},
 	"dependencies": {
-		"@automattic/generate-password": "^0.1.0"
+		"@automattic/generate-password": "^0.1.0",
+		"lodash": "^4.17.21"
```
iandunn: [Core already includes this](https://github.com/WordPress/WordPress/blob/1b393fbb15de13089b5c97a9c3975b642fbfbf7b/wp-includes/script-loader.php#L111), so bundling another version in our JS build might cause conflicts, and unnecessarily make the build larger.

Is there a tangible reason to add it, like the `import` statements breaking in some environments? Or was it added to document that it's used?

If it's the latter case, would it be better to mark it as being an global/external or something? IIRC there's a way to do that. 

We could also move it to `devDependencies` if it's needed for a tangible reason, though.

settings/src/components/password.js
```diff
@@ -26,11 +37,14 @@ import { GlobalContext } from '../script';
...
-	let passwordStrong                    = true; // Saved passwords have already passed the test.
+	const [ inputType, setInputType ] = useState( 'password' );
+	let passwordStrong = true; // Saved passwords have already passed the test.
 
 	if ( userRecord.hasEdits ) {
-		passwordStrong = isPasswordStrong( userRecord.editedRecord.password, userRecord.record );
+		passwordStrong = isPasswordStrong(
+			userRecord.editedRecord.password,
+			userRecord.record
+		);
```
iandunn, line 47: The linting/spelling changes are good, thanks! Ideally they'll be in a separate commit from the others (in `trunk`, not just this branch), so it's easier to see distinguish logic changes from linting. 

That's not a big deal though, so don't worry if the rebasing would take too long etc.

settings/src/components/password.js
```diff
@@ -3,18 +3,29 @@
...
+	useContext,
+	useEffect,
+	useState,
+} from '@wordpress/element';
+import {
+	Icon,
+	cancelCircleFilled,
+	check,
+	seen,
+	unseen,
```
iandunn, line 26: Great job adding the icon to Gutenberg!

I guess we'll need to wait until that's in a stable release to merge this?

settings/src/components/password.scss
```diff
@@ -1,19 +1,23 @@
...
-	}
 
 	.wporg-2fa__password_container {
 		justify-content: flex-start;
 		align-items: flex-start;
+		position: relative;
+		width: fit-content;
+
+		input {
+			padding-right: 35px;
```
iandunn, line 12:
```suggestion
			/* This includes room for the `(un)seen` icon, and also for the icon that 1password etc add to password fields. */
			padding-right: 35px;
```

iandunn (approved):
I left a few small comments, but this looks really good. Great job!

settings/package.json
outdoor2kode: > Or was it added to document that it's used?

Yeah, it was added because of linting error.

`10:1 error 'lodash' should be listed in the project's dependencies. Run 'npm i -S lodash' to add it  import/no-extraneous-dependencies`

> IIRC there's a way to do that.

Perhaps something like this [one](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/js/function-reference.js#L1)? 

Unfortunately, adding that comment doesn't fix the error, I've checked how Gutenberg handled it, and they just put it to `devDependencies`([ref](https://github.com/WordPress/gutenberg/blob/trunk/package.json#L205)) (The version is different from the core though, and they seem to [manually](https://github.com/WordPress/gutenberg/commit/d37bcb48d7028aa7701ff129ad6e338c65d1666f) updated it)

I've tried moving it to `devDependencies` and it solved the linting error, but I guess it's ok not to add it.

settings/src/components/password.js
outdoor2kode: Yeah, you're right. Will do it that way!

settings/src/components/password.js
outdoor2kode: If it is expected to be deployed soon after merging it, then it should wait until the GB 15.5 is released to be merged. Otherwise, it would cause errors.

outdoor2kode: Label: `[Status] On Hold` was added. I'll keep track of the Guteberg release schedule.

settings/package.json
iandunn: Ah, fixing the linting error makes sense 👍🏻 

It sounds like [webpack externals](https://webpack.js.org/configuration/externals/) might be a good way to do it. `wp-scripts` provides a way to [extend the default config](https://github.com/WordPress/gutenberg/tree/trunk/packages/scripts#provide-your-own-webpack-config). I'm not sure if the linter would pick up on that or not, though.

I think leaving it in `devDependencies` is a good solution too though 👍🏻

settings/src/components/password.js
iandunn: That sounds good 👍🏻  It is usually best to deploy soon after merging.

settings/package.json
StevenDufresne: I agree with the `devDependencies` approach.

StevenDufresne (approved):

PR #95, opened 2023-03-24
Add stub for WebAuthn UI

### LINKED ISSUE #87: Add WebAuthn UI stub
We should stub out the rough code and UI for the WebAuthn front-end, so we have something concrete to build off of. This is just the rough work, refinement will be done in #193 and #194.

- [x] this should be behind a flag so that production users don't see it until the backend is in place

WPCOM design as a reference, but it won't translate directly, and #194 will have a more refined version.

- [x] entry screen

this might not really fit into our design. we have the account status screen instead

Screenshot 2023-03-24 at 1 11 03 PM

- [x] naming key
Screenshot 2023-03-24 at 1 11 13 PM

- [x] spinner while waiting on browser to receive token
note: i don't think we should require a verification pin. it hurts the UX more than it helps security IMO

Screenshot 2023-03-24 at 1 11 46 PM

- [x] success notice
Screenshot 2023-03-24 at 1 13 51 PM

- [x] confirm removing key
Screenshot 2023-03-24 at 1 15 01 PM

iandunn:
Fixes #87 

This lays the rough foundation from a code perspective, but doesn't attempt to refine the UI yet. That'll be in subsequent PRs, in order to keep each one small enough to review easily. The adds the UI behind a feature flag, so that users won't see it until it's finished.

This will also give us something tangible to mock up the the UI, since the WPCOM one isn't a direct translation. @jasmussen , what do you think?

WebAuthn and TOTP disabled:

Screenshot 2023-05-24 at 3 05 59 PM

WebAuthn disabled, TOTP enabled:

Screenshot 2023-05-24 at 3 06 18 PM

WebAuthn enabled, TOTP disabled:

Screenshot 2023-05-24 at 3 06 34 PM

Initial WebAuthn screen:

Screenshot 2023-05-24 at 3 18 11 PM

Register a key:

Screenshot 2023-05-24 at 3 18 20 PM

Complete key registration:

Screenshot 2023-05-24 at 3 18 26 PM

_Note: #141 was broken off of this, but not merged. it probably doesn't have anything worth bringing back over, this is just an FYI_

### DIFF
--- a/settings/src/components/webauthn.js
+++ b/settings/src/components/webauthn.js
@@ -1,0 +1,199 @@
+/**
+ * WordPress dependencies
+ */
+import { Button, Spinner, TextControl } from '@wordpress/components';
+import { useCallback, useContext, useState } from '@wordpress/element';
+import { Icon, check } from '@wordpress/icons';
+
+/**
+ * Internal dependencies
+ */
+import { GlobalContext } from '../script';
+
+const confirm = window.confirm;
+
+/**
+ * Render the WebAuthn setting.
+ */
+export default function WebAuthn() {
+	const {
+		user: { userRecord },
+	} = useContext( GlobalContext );
+	const userKeys = userRecord.record[ '2fa_webauthn_keys' ];
+	const backupCodesEnabled =
+		userRecord.record[ '2fa_available_providers' ].includes( 'Two_Factor_Backup_Codes' );
+	const [ step, setStep ] = useState( 'manage' );
+
+	// todo this is just a placeholder. the real one would probably make an API request to save the new key,
+	// and then replace userkeys with the value returned by the call
+	// probably also refreshRecord( userRecord );
+	const onRegisterSuccess = useCallback( () => {
+		userKeys.push( {
+			id: 1111,
+			name: 'New Key',
+		} );
+		setStep( 'success' );
+
+		if ( ! backupCodesEnabled ) {
+			// todo redirect to backup codes
+		}
+	}, [ userKeys ] );
+
+	return (
+		<>
+			{ 'manage' === step && <Manage setStep={ setStep } userKeys={ userKeys } /> }
+
+			{ 'register' === step && (
+				<RegisterKey registerClickHandler={ () => setStep( 'waiting' ) } />
+			) }
+
+			{ 'waiting' === step && (
+				<WaitingForSecurityKey onRegisterSuccess={ onRegisterSuccess } />
+			) }
+
+			{ 'success' === step && <Success newKeyName={ 'Test key' } setStep={ setStep } /> }
+		</>
+	);
+}
+
+/**
+ * Render the Manage component.
+ *
+ * @param props
+ * @param props.setStep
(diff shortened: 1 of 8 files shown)


StevenDufresne:
Do you think we'll ever want to support the control of what method is "primary" if both are configured? If that is expected functionality, we should probably address it in this/series of PR(s). Maybe not necessarily the selection of the primary method, but indicating the order. Actually thinking more about it, we probably have to indicate somehow that webauth is the default method when both are configured anways.

settings/src/components/webauthn.js
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Render the WebAuthn setting.
+ */
+export default function WebAuthn() {
+	const [ step, setStep ] = useState( 'list' );
+
+	// TODO get keys from API.
+	const keys = [
+		{ name: 'Yubikey 5C Nano', id: 1 },
```
StevenDufresne: Could we get this working in this PR so its functionally complete?

jasmussen:
I can take a look at some design flows here when I get a moment. The screenshots are appreciated, video are as well.

iandunn:
>  we probably have to indicate somehow that webauthn is the default method when both are configured anyways.
 
Are you thinking something like this?

Screenshot 2023-05-25 at 2 08 17 PM

Or some changes to the backend?

Screenshot 2023-05-25 at 2 12 03 PM

settings/src/components/webauthn.js
iandunn: added in 0a4930a 👍🏻

iandunn:
Here's a gif of the rough flow, let me know if anything else would be helpful.

webauthn flow

StevenDufresne:
> > we probably have to indicate somehow that webauthn is the default method when both are configured anyways.
> 
> Are you thinking something like this?
> 
> Screenshot 2023-05-25 at 2 08 17 PM

On first glance, I missed the text added to the one time password list item. Maybe a nicer version of something like:

| Option | Option|
| --- | --- |
| Screenshot 2023-05-26 at 9 12 50 AM | Screenshot 2023-05-26 at 9 12 07 AM |

Maybe use the WebAuthn view to explain why it's default/primary (not sure what language is best.)?

iandunn:
Ah, yeah, that `Primary` one is much more obvious 👍🏻

jasmussen:
A quick take at some visuals. Since they, like last time, assume a future visual refresh, I will defer to you all on which parts to implememnt, and when. Hope this is helpful: a new checklist that uses color mainly to indicate what needs action.

checklist

Adding a security key flow:

security key

Note that I made this small little checkbox animation:

animation

I imagine it could be a fun little CSS component to animate, and it could be used when the security key is detected, then sit there for 2 seconds, and then automatically redirect you. Potentially we could also add it to similar flows, like successfully connecting 2FA and the likes.

iandunn:
Thanks! I opened #194 to track those. I want to keep this one to just the rough stubs, for easier iteration and focus.

PR #98, opened 2023-03-27
Setup improvements

outdoor2kode:
This PR
* Updates README
* Add pre-commit linter hooks

#### lint:js
image

#### lint:css
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ WordPress.org-specific customizations for the Two Factor plugin
 	}
 	add_action( 'init', __NAMESPACE__ . '\add_rewrite_rules' );
 	```
-1. Install, build, and activate the `wporg-support` theme.
+1. Install, build, and activate the [wporg-support](https://meta.svn.wordpress.org/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-support/.) theme.
 1. Install `bbPress` and `Gutenberg`. You might need to clone & build `trunk` branch of `Gutenberg` if we happen to be using any new features.
 1. `git clone` https://github.com/WordPress/two-factor/ into `wp-content/plugins` and follow their setup instructions.
 1. `git clone` this repo into `wp-content/plugins`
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -44,7 +44,7 @@ WordPress.org-specific customizations for the Two Factor plugin
 	}
 	add_action( 'init', __NAMESPACE__ . '\add_rewrite_rules' );
 	```
-1. Install, build, and activate the `wporg-support` theme.
+1. Install, build, and activate the [wporg-support](https://meta.svn.wordpress.org/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-support/.) theme.
```
iandunn:
```suggestion
1. Install, build, and activate the [wporg-support](https://meta.svn.wordpress.org/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-support/) theme.
```

settings/package.json
```diff
@@ -13,7 +13,8 @@
 		"packages-update": "wp-scripts packages-update",
 		"start": "wp-scripts start --webpack-copy-php",
 		"test:unit": "wp-scripts test-unit-js",
-		"test:unit:watch": "wp-scripts test-unit-js --watch"
+		"test:unit:watch": "wp-scripts test-unit-js --watch",
+		"prepare": "cd .. && husky install settings/.husky"
```
iandunn, line 17: Do we need to add a step in the `readme` that asks folks to run this, or does it get run automatically?

settings/.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
```
iandunn, line 1: I wonder if this would be better as a `pre-push` hook? I often make lots of WIP commits locally, and it can be annoying for the linter to run on those, since they're not intended to pass.

iandunn (approved):
This looks good 👍🏻 

It might be good to have the linter run on PRs too, but this is a good first step.

settings/.husky/pre-commit
outdoor2kode: It used to support pre-push, but the usage for the new version has changed, and pre-push doesn't seem to work anymore 🤔  I've tried for a while and also checked if others have encountered the same issue, but no good news. Let me do more research to see if I can find a solution.

settings/package.json
outdoor2kode: It would run automatically when `npm install` inside `settings`.

outdoor2kode:
> It might be good to have the linter run on PRs too, but this is a good first step.

I'll add it when I circle back to this PR later. I'll proceed to other higher-priority tasks at the moment, `pre-push` doesn't seem to work somehow on my side.

settings/.husky/pre-commit
iandunn: Ok, it's not a huge deal if we keep it `pre-commit`, especially if later on we add an action to do it on push.

settings/package.json
iandunn: Nice 👍🏻

iandunn:
That makes sense 👍🏻 The GH action isn't a high priority.

PR #99, opened 2023-03-27
Fix title position

### LINKED ISSUE #91: Title position is a bit off
The text `Password` is not in the middle.
image

outdoor2kode:
Fixes #91 

This PR fixes the position of the titles on the four main components.

## Screenshots

image
image
image
image

### DIFF
--- a/settings/src/script.js
+++ b/settings/src/script.js
@@ -128,7 +128,6 @@ function Main( { userId } ) {
 		screenContent = (
 			<Card>
 				<CardHeader className="wporg-2fa__navigation" size="xSmall">
-					<Flex>
 						<ScreenLink
 							screen="account-status"
 							anchorText={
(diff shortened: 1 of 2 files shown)


settings/src/script.js
```diff
@@ -128,7 +128,6 @@ function Main( { userId } ) {
 		screenContent = (
 			<Card>
 				<CardHeader className="wporg-2fa__navigation" size="xSmall">
-					<Flex>
```
iandunn, line 131: It looks like this was the only usage of `Flex`, so we can remove the `import` too.

iandunn (approved):
LGTM 👍🏻

settings/src/script.js
outdoor2kode: Ah, missed that, thanks.

PR #100, opened 2023-03-28
Split TOTP instruction into two steps

### LINKED ISSUE #93: Split 2FA instruction into two steps
The content in the 2fa screen is a bit packed together. We may want to split it into two steps. 

One step for the intro:
image

And one step for the non-QR code version:
image

Another for the text version:
image

outdoor2kode:
Fixes #93 

This PR
* Adds Intro component for totp introduction.
* Fixes the screen height of each step so when switching step, no jitter would happen.
* Adds AutoTabbingInput component
  * Only number can be submitted
  * Supports Auto Tabbing and 'backspace'
* Aligns styles with [design](https://www.figma.com/file/vmHxwa8WhiyiP9poe2om9J/Inspector-Panels?node-id=14404-127175&t=LlhNHIsuXe8pGf1E-0)

## Screencast

https://user-images.githubusercontent.com/18050944/228341459-09f68200-1857-415d-9d56-b25aea6d8d59.mp4

## TODO

* Lint after merging.
* Is there a better place for the `Error Notice`, or the current layout is acceptable? (screencast 0:40) @jasmussen @iandunn 

## Update

error


settings/src/components/totp.scss
```diff
@@ -14,22 +57,58 @@
...
+		background: $white !important;
+		box-shadow: (0 0 0 1px #3858E9) !important;
+		color: #3858E9 !important;
+	}
+
+	button {
+		background: #3858E9 !important;
+		border-color: #3858E9 !important;
+		color: $white !important;
+		margin-right: 0 !important;
```
outdoor2kode, line 89: Probably not a good idea to use that many `!important`, but as I don't want to build a button from scratch here, I just overrode the styles of Button from Gutenberg.

settings/src/components/auto-tabbing-input.scss
```diff
@@ -0,0 +1,17 @@
...
+		text-align: center;
+		border: 1px solid #949494;
+		border-radius: 2px;
+		padding: 0px !important;
+		
+		&:not(:last-child) {
+			margin-right: 24px;
+		}
+	}
+}
```
outdoor2kode, line 29: In this PR, there are still some files that haven't been well formatted yet. I will use the linter to fix them all at once later in a separate commit.

settings/src/components/totp.js
```diff
@@ -78,44 +116,39 @@ function Setup() {
...
+
+				<SetupForm
+					handleEnable={ handleEnable }
+					qrCodeUrl={ qrCodeUrl }
+					secretKey={ secretKey }
+					inputs={inputs}
+					setInputs={setInputs}
+				/>
+
+				<Button variant="link" onClick={ handleClick }>Previous</Button>
```
StevenDufresne, line 145: How does this differ from the "cancel" button under the 6 inputs?

StevenDufresne:
I don't find the introduction tab does _enough_ for it to warrant it's own tab to be honest. If it did a better job helping the setup, I would be interested in it, but since it just lists a few links, it seems like an extra interaction. I personally like how github does it:

Screenshot 2023-03-29 at 8 51 53 AM

I do very much like the input updates however. Does it make sense to split this up into those two features? One were we just update the code scan inputs and one that includes this extra "step"?

jasmussen:
Thanks for the ping, thank you for the video, that's useful as I'm not able to review the checked out branch. Thank you also for the attention to detail. 

First off, I think that _longer term_ we should move towards something like this.

Screenshot 2023-03-29 at 10 03 37

That is:

* A new notice component
* The new colors, including the blueberry and pomegrade.

For the near term, it looks a bit off to combine the blueberry with the classic WordPress blue. So you can think of the blueberry as a "spot color" — it should always be the spot color of the surrounding template. That surrounding template should eventually be updated to use blueberry, the sooner the better of course. But it's okay to keep the colors the same in the mean time. 

Same for the componentry — outside of an updated notice component, it's okay to use the existing notice component. Something like this:

Screenshot 2023-03-29 at 10 05 02

settings/src/components/totp.js
outdoor2kode: Oh, right! Originally, when the cancel was clicked, it would return to the account status screen. Replacing `previous` for that would make the look better. Thanks!

outdoor2kode:
> I do very much like the input updates however. Does it make sense to split this up into those two features? One were we just update the code scan inputs and one that includes this extra "step"?

It was suggested to splitting this up into two steps in that the original look was considered a bit packed together. I guess the main difference between us and Github is that we try to provide a few explanations and introductions about 2FA for .org here, whereas Github has [a whole section](https://docs.github.com/en/authentication/securing-your-account-with-two-factor-authentication-2fa/about-two-factor-authentication) dedicated to it. If we can provide more information on the introduction page, then that page may have more value. However, I personally feel that we can also simply provide the feature like Github does, without any introductory text in that maybe 2fa is common enough (maybe?) This way, we can have things on one page.

@jasmussen Do you have any thoughts on what @StevenDufresne brought up ^?
And thanks for the feedback and the clear clarification! I think I'll follow your idea that as the whole website is still in the classic WP style, and a new style may seem a bit abrupt, I'm not going to do extra adjustments going forward. However, I will add a red border when the input is wrong.

Regarding the repositioning of the notice, what's your thought on the things to see when there is no notice? 
I'm thinking it may look something like when there's no notice, the QR code and input fields are still kept close together, and once it appears, the input field is pushed down. If this is the case, an animation might be needed to help make this transition smoother.

jasmussen:
> Do you have any thoughts on what StevenDufresne brought up ^?

If the point is to combine two steps into one, I think that's fine. I'd suggest that also gives an opportunity to do a pass on the text, reduce when there's overlap. Less is always more when it comes to that. Doesn't have to be this PR, just noting.

> Regarding the repositioning of the notice, what's your thought on the things to see when there is no notice?
I'm thinking it may look something like when there's no notice, the QR code and input fields are still kept close together, and once it appears, the input field is pushed down. If this is the case, an animation might be needed to help make this transition smoother.

It's fine that the notice pushes content around 👍 👍

But especially if two steps are combined into one, it'd be good to reduce elements (such as text) as much as possible, so we still mostly retain the same footprint and shape of the overall box. If you can try it and add a screenshot, I can give some suggestions on that if helpful.

settings/src/components/totp.js
outdoor2kode: I found that it makes sense to me to use the `Cancel` button to clear all 6 inputs and keep the `previous` button.
Commit added: [Let Cancel clear all the inputs](https://github.com/WordPress/wporg-two-factor/pull/100/commits/cba6a73f03a3b18ed417154be8955049a8236bb2)

outdoor2kode:
I'm going to merge this first and https://github.com/WordPress/wporg-two-factor/issues/104 (New copy for TOTP), https://github.com/WordPress/wporg-two-factor/issues/105 (Notice reposition) were opened to address what we've discussed above.

PR #103, opened 2023-04-12
Encrypt TOTP keys

### LINKED ISSUE #101: Replace TOTP class with our own, to encrypt all the things
Using https://github.com/WordPress/two-factor/pull/546 we should replace the core TOTP class with our own instance of it and encrypt the secret that's being stored in the database.

See the above PR for an example of the class we could implement.

dd32:
Fixes #101

Depends upon https://github.com/WordPress/wporg-mu-plugins/pull/390
Made possible by https://github.com/WordPress/two-factor/pull/546

NOTE: if you test this on a sandbox (and then remove this), you will be unable to login with TOTP again without deactivating TOTP.

Testing Instructions:
 - Make sure you're running a version of `two-factor` with https://github.com/WordPress/two-factor/pull/546 merged
 - Define `WPORG_ENCRYPTION_KEY` and optionally `WPORG_TWO_FACTOR_ENCRYPTION_KEY`, with a value from `\WordPressdotorg\MU_Plugins\Encryption\generate_encryption_key()`
 - View `_two_factor_totp_key` value. Validate that it appears plaintext.
 - Login with TOTP.
 - View `_two_factor_totp_key` value. Validate that it appears to be encrypted (Long, and prefixed with `$t1$`).
 - Deactivate TOTP.
 - Setup TOTP.
 - Validate everything worked as expected.

TODO:
 - [x] Unit Tests.
 - [x] https://github.com/WordPress/wporg-mu-plugins/pull/390 (This is not required prior to this merging, but is required for it to actually encrypt)
 - [x] Test once 390 is merged
 - [x] Test on production sandbox, verify TOTPs are upgraded upon login.

### DIFF
--- a/tests/test-wporg-two-factor.php
+++ b/tests/test-wporg-two-factor.php
@@ -1,6 +1,7 @@
 <?php
 
 use function WordPressdotorg\Two_Factor\{ user_requires_2fa };
+use function WordPressdotorg\MU_Plugins\Encryption\{ generate_encryption_key };
 
 defined( 'WPINC' ) || die();
 
@@ -19,6 +20,17 @@ class Test_WPorg_Two_Factor extends WP_UnitTestCase {
 			'user_login' => 'regular_user',
 			'role'       => 'contributor',
 		) );
+
+		// Generate an encryption key for testing with.
+		if ( ! function_exists( 'wporg_encryption_keys' ) ) {
+			function wporg_encryption_keys() {
+				static $keys = null;
+
+				return $keys ?? $keys = [
+					'two-factor' => generate_encryption_key(),
+				];
+			}
+		}
 	}
 
 	/**
--- a/tests/test-wporg-two-factor.php
+++ b/tests/test-wporg-two-factor.php
@@ -230,10 +230,14 @@ class Test_WPorg_Two_Factor extends WP_UnitTestCase {
 		$enabled       = Two_Factor_Core::enable_provider_for_user( self::$regular_user->ID, 'Two_Factor_Totp' );
 		$this->assertTrue( $enabled );
 
-		// Validate that the TOTP key was stored in an encrypted form.
-		$totp_key       = $totp_provider->get_user_totp_key( self::$regular_user->ID );
-		$totp_user_meta = get_user_meta( self::$regular_user->ID, $totp_provider::SECRET_META_KEY, true );
-		$this->assertNotSame( $totp_key, $totp_user_meta );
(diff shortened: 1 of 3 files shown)


tests/test-wporg-two-factor.php
```diff
@@ -229,7 +229,7 @@ public function test_set_primary_provider_for_user() {
 		$totp_provider->set_user_totp_key( self::$regular_user->ID, Two_Factor_Totp::generate_key() );
 		$enabled = Two_Factor_Core::enable_provider_for_user( self::$regular_user->ID, 'Two_Factor_Totp' );
 
-		$expected = 'Two_Factor_Totp';
+		$expected = 'WordPressdotorg\Two_Factor\Encrypted_Totp_Provider';
 		$actual   = get_class( Two_Factor_Core::get_primary_provider_for_user( self::$regular_user->ID ) );
 		$this->assertTrue( $enabled );
 		$this->assertSame( $expected, $actual );
```
dd32: This test can probably change to both check the class, but also check it's Totp as expected, as that's what we set as active above.

```suggestion
		$this->assertTrue( $enabled );

		$provider = Two_Factor_Core::get_primary_provider_for_user( self::$regular_user->ID );

		$expected_class = 'WordPressdotorg\Two_Factor\Encrypted_Totp_Provider';
		$actual_class   = get_class( $provider );
		$this->assertSame( $expected_class, $actual_class );

		$expected_key = 'Two_Factor_Totp';
		$actual_key   = $provider->get_key();
		$this->assertSame( $expected_key, $provider->get_key() );
```

pkevan: Good point, I was primarily fixing the tests before testing - makes sense to cover both.

pkevan (approved):

pkevan:
Tested on a sandbox, and confirmed the TOTP key is encrypted, and you can log in once it has been encrypted.

dd32:
I've updated this to include the User ID as part of the encryption, such that user_meta for one user is invalid for another.
_(I've 'upgraded' the existing encrypted TOTP keys on WordPress.org)._

I've updated the unit tests to test that the key is stored in a format different than the key (ie. encrypted) but left the validation of the encryption functions to the Encryption library PR.
_edit: Of course this doesn't actually pass yet.. as the encryption methods aren't merged.. [3d89b20](https://github.com/WordPress/wporg-two-factor/pull/103/commits/3d89b204f8aa99bfb12d1e509c7479179946f126) can be reverted once the upstream PR is merged._

Issue #108, opened 2023-04-18
Scan QR Code "back" navigation isn't clear.


StevenDufresne:
The cancel in relation to the "previous" doesn't make sense. I think we need to do one of the following:

1. Change "cancel" to "clear", move the "previous" button on the left of the UI
2. Remove the "cancel" button, move the "previous" button on the left of the UI

### Screenshot
<img width="450" alt="Screenshot 2023-04-18 at 1 44 36 PM" src="https://user-images.githubusercontent.com/1657336/232674667-c701e1dc-76bc-42fd-94bc-63e70fe8630d.png">

outdoor2kode:
Fixed in https://github.com/WordPress/wporg-two-factor/pull/112

Issue #105, opened 2023-04-13
Notice reposition


outdoor2kode:
Animation is needed - The notice would push down the 6-digit inputs below.

![image](https://user-images.githubusercontent.com/18050944/231714398-75d8d9f9-b605-410f-a064-67ce59fe5069.png)

outdoor2kode:
Fixed in https://github.com/WordPress/wporg-two-factor/pull/112

PR #112, opened 2023-04-18
New copy for TOTP + Notice reposition

### LINKED ISSUE #104: New copy for TOTP
See discussion https://github.com/WordPress/wporg-two-factor/pull/100#issuecomment-1487757380.

We need a new copy for TOTP. If we'd like to keep it as two steps, the copy should be richer in the intro screen. Otherwise, the text needs to be reduced. ([ref](https://github.com/WordPress/wporg-two-factor/pull/100#issuecomment-1489823119))

outdoor2kode:
Fixes #104 #105 #108 

This PR 
1. Updated the copy and combined the TOTP steps into a single step.
2. Moved the Error Notice to a new position.
3. Changed "cancel" to "clear", and moved the "previous" button on the left of the UI.

## Screencast

https://user-images.githubusercontent.com/18050944/232840205-c8dcb4df-4a5e-42b6-86dc-9a49ddf10408.mp4

## Testing Steps

[Figma](https://www.figma.com/file/vmHxwa8WhiyiP9poe2om9J/Inspector-Panels?node-id=14591-120142&t=SjIDl48SpGin2L1W-0)
1. Go to `Two-Factor Authentication`.
2. Check if the left button is with the text `Clear`
3. Check if the setup instruction copy is updated.
4. Try some random digits.
5. Error notice should appear in the middle with an animation effect after submitting.
6. Digits, red borders, and the error notice should disappear after clicking `Clear`.
7. Try the same steps in manual mode.


outdoor2kode:
@jasmussen Could you help check the screencast in the PR description? I've centered the `Scan QR code` and its instruction `use an authenticator app...` as it seems a bit off when switching to manual mode if putting it on the left (see video below). What do you think?

https://user-images.githubusercontent.com/18050944/232962583-b81b86b6-6d69-4ef7-884d-27772b84930a.mp4

jasmussen:
Yes, thanks for bringing that to my attention, and appreciate all the energy you put into this.

Revisiting this, it's probably best to include the whole frame, including the header, the back button, and the stepper. If we rephrase slightly, remove a subheading, and remove the stepper, we can simplify it a fair bit. Shown here in the newer colors:

qr, near term

Here shown in existing colors:

qr

The above boxes the main QR / time code unit into a 360px max-width box that's centered horizontally. What do you think?

outdoor2kode:
Thanks @jasmussen, the new look looks better 👍 

One thing that would come up when removing the bottom line from the header: Other components are likely to also need some updates for their layout. Examples:
image
image

Does it make sense that we still keep the bottom line? (bolded `Setup authenticator app` and setup progress bar were removed)
Screen Capture on 2023-04-20 at 00-13-07

settings/src/components/totp.js
```diff
@@ -116,22 +79,33 @@ function Setup( { setIsNextClick } ) {
...
 
-	const handleClick = useCallback( () => setIsNextClick( false ), [] );
-
 	return (
 		<>
-			<SetupProgressBar step="totp-setup" />
+			<Flex expanded={false} direction='column' align="top" justify="top" gap="14px" className="wporg-2fa__totp_setup-container">
+				<strong>
+					Setup authenticator app
+				</strong>
```
iandunn: What do you think about using an `<h3>` for this (and similar instances)? It seems like that might be more semantic.

jasmussen:
Thanks for your patience. I think I'll need a moment in my morning to create mockups for those other cases, I'll provide you with that. 

But if it helps, I don't find it problematic to keep the bottom line on the header. So if that's easiest for you to go with now, definitely keep it simple.

settings/src/components/totp.js
```diff
@@ -249,18 +211,27 @@ function SetupForm( { handleEnable, qrCodeUrl, secretKey, inputs, setInputs, err
...
-				<Button variant="secondary" onClick={ handleCancelClick }>
-					Cancel
-				</Button>
-				<Button type="submit" variant="primary" disabled={ ! canSubmit }>
-					Enable
-				</Button>
-			</div>
-		</form>
+		<Flex expanded={false} direction='column' align='center' gap="16px" className="wporg-2fa__setup-form-container">
+			<Notice status="error" isDismissible={ false } className={ error ? "show" : ""}>
```
iandunn:
```suggestion
			<Notice status="error" isDismissible={ false } className={ error ? "is-shown" : ""}>
```

I think this might be more consistent with [Gutenberg naming standards](https://developer.wordpress.org/block-editor/contributors/code/coding-guidelines/#naming).

iandunn (approved):
LGTM 👍🏻

outdoor2kode:
Thanks @jasmussen, I think I will keep the bottom line and these suggestions - `rephrase slightly, remove a subheading, and remove the stepper`, it would make the look good enough to go.

settings/src/components/totp.js
outdoor2kode: Yeah, that makes sense, originally it was in `<p>` tag. This one will be removed but I'll review other similar instances.

settings/src/components/totp.js
outdoor2kode: Thanks 👍

outdoor2kode:
**Update video**

https://user-images.githubusercontent.com/18050944/233179156-be313330-cd78-4f95-9c36-4e2a541e17dd.mp4

StevenDufresne:
Mobile doesn't look quite right:

Screenshot 2023-04-20 at 10 35 33 AM

jasmussen:
Cool, let me know if you need anything else.

PR #113, opened 2023-04-18
Make users profile page look less tight

### LINKED ISSUE #102: Add padding below the 2fa control.
Not sure where to log or make this update, but we should add more padding below our component as the page feels pretty tight.

| Before | After | 
| --- | --- |
| localhost_8888_users_test_edit_account__screen=account-status (1) |  localhost_8888_users_test_edit_account__screen=account-status |

outdoor2kode:
Fixes #102 

This PR
1. Removes height restriction that makes the layout go off.
2. Add `margin-bottom: 2em` to the profile page for a less crowded look.
   a. `2em` comes from the `margin-top` value set in `#main`.

## Screencast

https://user-images.githubusercontent.com/18050944/232853625-f677a9bf-22d5-4422-8050-069092eb6d07.mp4

### DIFF
--- a/settings/src/style.scss
+++ b/settings/src/style.scss
@@ -53,11 +53,6 @@
 	letter-spacing: .3em;
 }
 
-.wporg-2fa__screen-wrapper {
-	min-height: 665px !important;
-    height: unset !important;
-}
-
 .components-notice {
 	margin-left: 0;
 	margin-right: 0;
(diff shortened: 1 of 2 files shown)


settings/src/style.scss
```diff
@@ -1,6 +1,10 @@
 @import '~@wordpress/base-styles/colors';
 @import '~@wordpress/base-styles/colors.native';
 
+#main {
+	margin-bottom: 2rem;
+}
```
StevenDufresne, line 6: I _think_ this makes sense. Just double check that it will work in your sandbox once this code is in place.

StevenDufresne (approved):

PR #123, opened 2023-04-20
Clear notice when inputs array is empty on TOTP screen

### LINKED ISSUE #106: Scan QR Code - Clearing input should remove error state.
When the inputs are cleared, the error state should  be cleared.

## Screenshot
Screenshot 2023-04-18 at 1 34 48 PM

## Test
1. Enter random digits
2. Notice all the fields go red
3. Clear all inputs, notice they are still red.

outdoor2kode:
Fixes #106 

## Screenshot

clear error notice

### DIFF
--- a/settings/src/components/totp.js
+++ b/settings/src/components/totp.js
@@ -189,7 +189,7 @@ function SetupForm( { handleEnable, qrCodeUrl, secretKey, inputs, setInputs, err
 	const [ isInputComplete, setIsInputComplete ] = useState(false);
 
 	const handleComplete = useCallback( ( isComplete ) => setIsInputComplete( isComplete ), [])
-	const handleCancelClick = useCallback( () => { 
+	const handleClearClick = useCallback( () => { 
 		setInputs( Array( 6 ).fill( '' ) );
 		setError('');
 	}, [])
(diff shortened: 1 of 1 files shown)


StevenDufresne:
Does it make sense to clear the error state once they remove the last value?

outdoor2kode:
> Does it make sense to clear the error state once they remove the last value?

We could, but it's probably that someone somehow deletes the first digit first, likely an edge case, though.

If we're looking to not iterate the whole array every time, logic like `! inputs.some( input => input !== '')` would work, but it's a bit convoluted. And I guess it's ok to iterate for only 6 digits with only entering the condition when there's an error.

StevenDufresne:
> We could, but it's probably that someone somehow deletes the first digit first, likely an edge case, though.

Not the first digit, the last one, so as the user is reversing their changes, as soon as we get to:

```
[ 1 ] [ 2 ] [ 3 ] [ 4 ] [ 5 ] [ 7 ] <-- error shown
[ 1 ] [ 2 ] [ 3 ] [ 4 ] [ 5 ] [   ] <-- remove error state
[ 1 ] [ 2 ] [ 3 ] [ 4 ] [   ] [   ]
...
```

remove the error messages since it's technically not incorrect anymore.

outdoor2kode:
> Not the first digit, the last one, so as the user is reversing their changes, as soon as we get to:

Ah, thanks for the clarification! Sorry for not fully understanding your idea at the beginning.

**Update Gif**

Screen Capture on 2023-04-21 at 11-30-41

StevenDufresne (approved):
I _think_ this works. 

More people using 2fa will help inform this better so I vote we merge this and ask for more general feedback later.

PR #124, opened 2023-04-20
Use createRoot from wp.element

outdoor2kode:
Happen to see this while doing a11y audit, the PR has landed in GB 15.1, so we can use createRoot from there now.


iandunn (approved):
Looks good, thanks for noticing it and checking!

PR #125, opened 2023-04-20
A11y audit

### LINKED ISSUE #110: A11y Audit
**Keyboard**

- [x] Interactive controls are keyboard focusable.
    - [x] Reasonable DOM order.
    - [x] Non-interactive content should avoid focusable.
- [x] User focus is not accidentally trapped in a region.

**Screen Reader**

- [x] Custom controls have ARIA roles.
- [x] Offscreen content is hidden from assistive technology.
- [x] Clear page structure. (h1-h3 hierarchy is clear)
~HTML5 landmark elements are used to improve navigation.~

**Interactive elements should be distinguishable**

- [x] Ensure link `WordPress/wporg-two-factor on GitHub` is distinguished from surrounding text in a way that does not rely on color.

outdoor2kode:
Fixes #110 

This PR tested items listed in #110 and did changes accordingly.

For `HTML5 landmark elements are used to improve navigation`, originally, I was trying to add `role="list"` or `role="region"` to the setting list in the `Main` section to improve screen reader navigation, but I found that VoiceOver couldn't detect either of them. Among the other roles, though it can detect `role="article"`, it doesn't seem to be a very suitable role. So I didn't make any changes to that item.

Screen Shot 2023-04-20 at 10 00 05 PM

Screen Shot 2023-04-20 at 9 59 35 PM

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -44,7 +44,7 @@ function render_custom_ui() : void {
 		return;
 	}
 
-	echo '<div class="notice notice-info notice-alt"><p>This interface and functionality is currently a <em>beta release</em>, while it is currently operational it\'s not yet officially released. Please report any issues you encounter to <a href="https://github.com/WordPress/wporg-two-factor/issues" data-content="notice-info-link">WordPress/wporg-two-factor on GitHub</a>.</p></div>';
+	echo '<div class="notice notice-info notice-alt"><p>This interface and functionality is currently a <em>beta release</em>, while it is currently operational it\'s not yet officially released. Please report any issues you encounter to <a href="https://github.com/WordPress/wporg-two-factor/issues">WordPress/wporg-two-factor on GitHub</a>.</p></div>';
 
 	$user_id    = bbp_get_displayed_user_id();
 	$json_attrs = json_encode( [ 'userId' => $user_id ] );
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -44,7 +44,7 @@ function render_custom_ui() : void {
 		return;
 	}
 
-	echo '<div class="notice notice-info notice-alt"><p>This interface and functionality is currently a <em>beta release</em>, while it is currently operational it\'s not yet officially released. Please report any issues you encounter to <a href="https://github.com/WordPress/wporg-two-factor/issues">WordPress/wporg-two-factor on GitHub</a>.</p></div>';
+	echo '<div class="notice notice-info notice-alt"><p>This interface and functionality is currently a <em>beta release</em>, while it is currently operational it\'s not yet officially released. Please report any issues you encounter to <a href="https://github.com/WordPress/wporg-two-factor/issues" data-content="notice-info-link">WordPress/wporg-two-factor on GitHub</a>.</p></div>';
 
 	$user_id    = bbp_get_displayed_user_id();
 	$json_attrs = json_encode( [ 'userId' => $user_id ] );
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -73,7 +73,7 @@ function SettingStatusCard( { screen, status, headerText, bodyText } ) {
 				anchorText={
 					<CardBody>
 						<StatusIcon status={ status } />
-						<h3>{ headerText }</h3>
+						<h3 aria-label={headerText + '.'}>{ headerText }</h3>
 						<p>{ bodyText }</p>
 						<Icon icon={ chevronRight } size={ 26 } className="wporg-2fa__status-card-open" />
 					</CardBody>
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -73,7 +73,7 @@ function SettingStatusCard( { screen, status, headerText, bodyText } ) {
 				anchorText={
 					<CardBody>
 						<StatusIcon status={ status } />
-						<h3 aria-label={headerText + '.'}>{ headerText }</h3>
+						<h3 aria-label={"Click to enter the " + headerText + ' setting page.'}>{ headerText }</h3>
 						<p>{ bodyText }</p>
 						<Icon icon={ chevronRight } size={ 26 } className="wporg-2fa__status-card-open" />
 					</CardBody>
--- a/settings/src/components/email-address.js
+++ b/settings/src/components/email-address.js
@@ -91,7 +91,7 @@ export default function EmailAddress() {
 				help="We will send you a verification email before updating your email address."
 				label="Your email address"
 				size="62"
-				placeholder="my-email-address@example.org"
+				placeholder="Enter email address..."
 				value={ editedRecord.email ?? record.email }
 				onChange={ (email) => edit( { email } ) }
 			/>
--- a/settings/src/style.scss
+++ b/settings/src/style.scss
@@ -5,8 +5,13 @@
 	margin-bottom: 2rem;
 }
 
-a[data-content*="notice-info-link"] {
-	text-decoration: underline;
+#bbp-your-profile {
+	.components-notice,
+	.notice {
+		a {
+			text-decoration: underline;
+		}
+	}
 }
 
 .wp-block-wporg-two-factor-settings {
--- a/settings/src/style.scss
+++ b/settings/src/style.scss
@@ -5,6 +5,10 @@
 	margin-bottom: 2rem;
 }
 
+a[data-content*="notice-info-link"] {
+	text-decoration: underline;
+}
+
 .wp-block-wporg-two-factor-settings {
 	position: relative;
 
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -128,7 +128,7 @@ export default function Password() {
 					label="New Password"
 					size="62"
 					value={ userRecord.editedRecord.password ?? '' }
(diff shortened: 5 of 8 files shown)


settings/src/style.scss
```diff
@@ -5,6 +5,10 @@
 	margin-bottom: 2rem;
 }
 
+a[data-content*="notice-info-link"] {
+	text-decoration: underline;
+}
+
```
iandunn: What do you think about applying this to all `.notice` links? it seems like error/warning notices would have the same issue as info.

settings/settings.php
```diff
@@ -44,7 +44,7 @@ function render_custom_ui() : void {
 		return;
 	}
 
-	echo '<div class="notice notice-info notice-alt"><p>This interface and functionality is currently a <em>beta release</em>, while it is currently operational it\'s not yet officially released. Please report any issues you encounter to <a href="https://github.com/WordPress/wporg-two-factor/issues">WordPress/wporg-two-factor on GitHub</a>.</p></div>';
+	echo '<div class="notice notice-info notice-alt"><p>This interface and functionality is currently a <em>beta release</em>, while it is currently operational it\'s not yet officially released. Please report any issues you encounter to <a href="https://github.com/WordPress/wporg-two-factor/issues" data-content="notice-info-link">WordPress/wporg-two-factor on GitHub</a>.</p></div>';
```
iandunn: What do you like about using `data-content` over `class`? `class` seems more conventional to me, but I'm curious if there's a benefit to `data-content` that I'm not aware of.

settings/src/components/account-status.js
```diff
@@ -73,7 +73,7 @@ function SettingStatusCard( { screen, status, headerText, bodyText } ) {
 				anchorText={
 					<CardBody>
 						<StatusIcon status={ status } />
-						<h3>{ headerText }</h3>
+						<h3 aria-label={"Click to enter the " + headerText + ' setting page.'}>{ headerText }</h3>
```
iandunn, line 76: 🤔 Is the label needed because the screen reader doesn't know that it's a "button". If so, would `aria-role=button` be a better fit? Having `Click to...` in a label seems like it might be a "smell" for similar reasons that ["click here" is bad link text](https://mrwweb.com/think-before-you-link/). If we can rely on semantics then that seems better to me.

If you do think `aria-role` is a good idea, would it be better to place it in `ScreenLink`, so that it doesn't have to be duplicated in each instance?

settings/src/components/email-address.js
```diff
@@ -91,7 +91,7 @@ export default function EmailAddress() {
 				help="We will send you a verification email before updating your email address."
 				label="Your email address"
 				size="62"
-				placeholder="my-email-address@example.org"
+				placeholder="Enter email address..."
```
iandunn, line 94: My understanding is that [`placeholder` is intended to be used for examples, rather than explanations](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#placeholder). The existing `label` field seems like it acts as an explanation, and I think having 1 explanation and 1 example is more helpful than 2 explanations.

Is the problem that the `label` isn't descriptive enough? If so, would improving it be better for screen readers and sighted users? Or maybe adding an `aria-label` would be better than removing the example?

I could definitely be wrong, though, so I'm curious what you think.

( _this applies to the other `placeholder` explanations too._ )

settings/src/components/account-status.js
iandunn, line 76: Should this use the new `ariaLabel` parameter for `ScreenLink`, or is it better to have the label on the `h3` here, but on the `a` in other places?

Was it placed here to add the pause from 1b6a8adc6b7ad1cfb24a657ec97b83cee55b40f4? I wonder why [the screen reader doesn't pause automatically](https://stackoverflow.com/a/15883990/450127), since it knows that this is a header and the following is a paragraph.

iandunn (commented):
I think this is looking good, but I have a few questions about some parts that I'm not sure about.

settings/settings.php
outdoor2kode: I believe the use of `data-content` originally came from trying to find a solution to the issue `Screen reader stops reading paragraph when it reaches anchor tag`, but in the end, I found this actually is an accessible-friedly way: It gives the user the opportunity to follow the links when the screenreader reaches them and waits before it proceeds. `class` should be used here. Thanks for pointing out 🙏

settings/src/style.scss
outdoor2kode: Since the other two don't have `<a>` on them, and I assumed this issue would only happen with the "info" noitce (where the contrast ratio of `text : background color` and `text : surrounding text color` are not sufficient), so I didn't specifically think about handling it. But it also makes sense to apply this to all `.notice` links I believe 👍

settings/src/components/account-status.js
outdoor2kode: > Should this use the new ariaLabel parameter for ScreenLink, or is it better to have the label on the h3 here, but on the a in other places?

Currently, if the label is on the h3 here, both text in `h3` and `p` would be read in a screen reader, otherwise, only `Click to enter the password setting page.` would be read if using `ariaLabel` of ScreenLink.

image

> I wonder why [the screen reader doesn't pause automatically](https://stackoverflow.com/a/15883990/450127), since it knows that this is a header and the following is a paragraph.

I guess the reason is If you put the h3 and p tags inside an a tag, the behavior of screen readers, like VoiceOver I'm using for testing at the moment, might treat the content of the h3 and p tags as part of a single link, so there might not be a significant pause between them while being read.

And typically, to maintain good accessibility and clear structure, it's better to place the a tag inside the h3 or p tags, rather than having them as children of the a tag. For example:
```
<h3><a href="#">{ headerText }</a></h3>
<p>{ bodyText }</p>
```
but in our case, we could also change the `a` to `div`  [here](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/screen-link.js#L27) to make a screen reader read h3 and p separately. It would require a few extra CSS tweaks for `ScreenLink`, though (and adding an [aria-role](https://github.com/WordPress/wporg-two-factor/pull/125#discussion_r1175632761) as well).
image

outdoor2kode: Initially, I added a label to add a [pause](https://github.com/WordPress/wporg-two-factor/commit/1b6a8adc6b7ad1cfb24a657ec97b83cee55b40f4) between `<h3>` and `<p>`. Later on, I was thinking to provide a clearer instruction by adding "Click to..." to it. I'm not sure if this is also a smell because in the example of "click here" that you brought up, it's true that it doesn't provide any useful information to people who can visually see the link. But in this case, I was thinking, just saying "Password" might not be clear enough for people with visual impairments. But, I might be overthinking it because as seen in the image below, when a screen reader reaches the end of the text, it still says "You are currently on a link. To click this link, press...". So, people who use screen readers may not need the additional "Click to..." because they are probably already accustomed to hearing "You are currently on a link" at the end. 

image
image

settings/src/components/email-address.js
outdoor2kode: image

The reason I made the change here was that since we used `TextControl` to generate the `label` and `input`, and it put the placeholder within the input, this would cause the screen reader to read the placeholder when reading the input. Normally, there are two ways to avoid the placeholder from being read by the screen reader, but neither can be done in this case.

```
<label for="password">Password: 
    <span aria-hidden="true">@aasd#%RGG$GG@#R</span>
</label>
<input type="password" id="password" name="password" placeholder="">
```
or 
```
<label for="password">Password:</label>
<input type="password" id="password" name="password" aria-labelledby="passwordExample">
<div id="passwordExample" aria-hidden="true">@aasd#%RGG$GG@#R</div>
```

And If we still want the placeholder to be used for an example, we would need to add an `aria-label` to override the `label`, otherwise the screen reader would read the label + placeholder text without pausing (like the img above shows). And the way to override the `label` is kind of tricky.

```
label="New Password"
aria-label=" " // This can prevent the screen reader from reading the "New Password" label and the placeholder text without any pauses.
placeholder="lsfijklewja2#@$@#$slkfj"
```
image

settings/src/style.scss
outdoor2kode: Fixed in https://github.com/WordPress/wporg-two-factor/pull/125/commits/1ef8591a093b3110d5dbb54d274bc55d1b003122

settings/settings.php
outdoor2kode: Fixed in https://github.com/WordPress/wporg-two-factor/pull/125/commits/1ef8591a093b3110d5dbb54d274bc55d1b003122

settings/src/components/account-status.js
iandunn: 🤔 that's a good point. I'm not sure, so I'd say go with whichever you think is best, and we can always iterate if we get any feedback about it 👍🏻

settings/src/components/email-address.js
iandunn: That works for me, thanks for the explanation 👍🏻

settings/src/components/account-status.js
iandunn: That makes sense. I think we can leave it the way you've done it 👍🏻

iandunn (approved):
LGTM, great work 👍🏻

PR #129, opened 2023-04-22
Show password errors on submit to improve UX

### LINKED ISSUE #118: Show password strength indicator only submit & failure
We currently show the error message as soon as a single character is entered into the password field and even when the password field is empty. This is distracting for a user who is typing a password.

My initial reaction is that, we should probably only show on submit until we have a better inline solution for password feedback. 
We could also throttle the callback so it's only called after we _think_ the user has completed typing their password or is closer to.

Other opinions?

### GIF
Screen Capture on 2023-04-20 at 10-37-25

iandunn:
Fixes #118

This only shows a password error after the user has tried to submit the form. Before that, showing it can be annoying because they know it's not valid until they've finished trying. After they've attempted to save, though, it can be helpful to get instant acknowledgement when the password is strong enough.

This doesn't change behavior of "is strong enough" notice at all. That seems good to me, but I don't feel strongly.

password ux

### DIFF
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -165,11 +165,7 @@ export default function Password() {
 			<p>
 				<Button
 					isPrimary
-					disabled={
-						passwordStrong && ! userRecord.isSaving
-							? ''
-							: 'disabled'
-					}
+					disabled={ ! userRecord.editedRecord.password }
 					type="submit"
 				>
 					{ userRecord.isSaving ? 'Saving...' : 'Save password' }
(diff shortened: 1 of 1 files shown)


StevenDufresne (approved):
This is definitely an improvement. The implementation also makes sense.

I think the best experience is to have an inline UI since we do have password strength requirements. I'll open a ticket for that feature request. [Edit [ticket](https://github.com/WordPress/wporg-two-factor/issues/131)]

PR #132, opened 2023-04-24
Add yarn workspace.

### LINKED ISSUE #130: Create a Build branch / action
Similar to other WordPress.org GitHub repo's, we should create a `build` branch which is pulled from for deployment to WordPress.org.

This allows us to run the build steps within a GitHub Action, isolated from the configuration of the WordPress.org deploy system.

Why? The deployed build broke via the deploy system, causing the build folder to be removed without notice, while that was user-error not noticing the deletion, we should remove as much human error as possible.

StevenDufresne:
Relates to: #130 

This adds the settings folder as a workspace so we can run it from the project root, making developer experience better, bringing this project more aligned with other code bases. 

This PR will be the base for the GH action that fixes #130


package.json
```diff
@@ -5,9 +5,9 @@
   "directories": {
     "test": "tests"
   },
+  "private": true,
   "scripts": {
-    "env": "wp-env",
-    "postinstall": "cd settings && npm install --force && cd .. && wp-env start && wp-env run composer install",
```
dd32, line 10: I'm wondering if this should remain, but just as the composer install? 

```
"postinstall": "wp-env run composer install",
```

dd32 (approved):
This approach makes sense to me.

package.json
StevenDufresne: I was going to note this. Why is it needed?

StevenDufresne: Nvm, appears related to the test framework.

PR #134, opened 2023-04-25
Fork upstream WebAuthn provider

### LINKED ISSUE #114: Turn on WebAuthn in wp-admin
For the MVP, let's turn on WebAuthn in the upstream plugin and verify that it works. If it does, I think we can launch with support for privileged users and iterate, adding support for all users. 

- [x] Decide which approach to take
- [x] ~~Port MadWizard to upstream PR~~
    - [ ] #134 
    - [ ] #146
    - [ ] Port login and test flows to use MadWizard library
    - [ ] Cleanup and improvements
    - [ ] Holistic code review now that the finished piece can all be seen together, rather than smaller PRs where much of the code is temporary
- [x] #153 
- [x] https://github.com/WordPress/wporg-two-factor/pull/165
- [x] https://github.com/WordPress/wporg-two-factor/pull/169
- [x] ~~Update #141 to show if enabled/disabled state programmatically.~~
- [ ] Get security review and iterate
- [x] ~~Enable the link to wp-admin from #141~~

iandunn:
See #114

This forks the provider from https://github.com/WordPress/two-factor/pull/427/ as starting point for our customized version of it. More work will be needed to close that issue, but I'm planning to do it across several PRs to keep it manageable.

This one just forks the upstream PR and gets it working inside our plugin. Future PRs will migrate to the MadWizard library, remove abstractions, increase test coverage, and make other changes to fit w.org's needs. After we refine everything, we can eventually contribute it upstream as an alternate proposal.

A lot of this will be removed or changed significantly in the following PRs, so the only parts that are really worth reviewing are:

* `phpunit.xml.dist`
* `wporg-two-factor.php`
* `providers/webauthn/class-two-factor-webauthn.php`
* `providers/webauthn/webauthn-admin.js`
* `providers/webauthn/webauthn-login.js`
* `providers/webauthn/webauthn-login.js`

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ WordPress.org-specific customizations for the Two Factor plugin
 		return in_array( $user->user_login, $GLOBALS['supes'], true );
 	}
 	```
-1. Install and build the `wporg-mu-plugins` repository.
+1. `git clone https://github.com/WordPress/wporg-mu-plugins.git --branch build wp-content/mu-plugins/pub`
 1. Add this code to your `wp-content/mu-plugins/0-sandbox.php`:
 	```php
 	require_once WPMU_PLUGIN_DIR. '/wporg-mu-plugins/mu-plugins/loader.php';
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -30,6 +30,7 @@
 	  -->
 	<coverage processUncoveredFiles="false">
 		<include>
+			<directory>./providers</directory>
 			<file>wporg-two-factor.php</file>
 			<file>class-encrypted-totp-provider.php</file>
 			<file>settings/settings.php</file>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -31,6 +31,7 @@
 	<coverage processUncoveredFiles="false">
 		<include>
 			<file>wporg-two-factor.php</file>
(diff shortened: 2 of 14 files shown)


pkevan:
Anything in particular we should be looking for or paying attention to. I don't have a suitable device to test with, but unsure if that is necessary?

iandunn:
I tested by registering 2 keys, making sure the "test" button worked for each, and that I could log in with each, and then deleting the keys. I also tested the interim login modal.

IMO the most important thing at this phase is identifying any significant code changes that we'll want to make to the provider itself (not its libraries). If you look through the code in the files mentioned above, and note any red flags, that'd be really helpful!

iandunn:
You may be able to [use a software key to test with](https://stackoverflow.com/questions/52445624/how-to-use-webauthn-without-key-fob), but I'd recommend expensing a Yubikey or something similar.

pkevan:
> Yubikey or something similar.

Should have one tomorrow, so will be able to test this then.

iandunn:
Moving this back to a draft because integrating this and #146 is turning out to be more time consuming that I hoped. I'm going to [try another approach instead](https://github.com/WordPress/wporg-two-factor/issues/114#issuecomment-1533642960).

pkevan:
I didn't test this PR, as it's moved back to draft - had a quick skim of the code and nothing stood out.

PR #136, opened 2023-04-25
Clear incomplete state when navigating away page

### LINKED ISSUE #117: Clear incomplete states when a user navigates away from a view.
When a users starts changing their password, abandons and navigates away, when they return, the state has not changed. Should we clear the state on backwards navigation?

## GIF

Screen Capture on 2023-04-20 at 10-32-18

outdoor2kode:
Fixes #117

This PR clears the incomplete state when navigating away from the password setting page.

password

### DIFF
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -40,6 +40,9 @@ export default function Password() {
 	const [ inputType, setInputType ] = useState( 'password' );
 	const [ hasAttemptedSave, setHasAttemptedSave ] = useState( false );
 	let passwordStrong = true; // Saved passwords have already passed the test.
+	// Originally, password wasn't set in record, but we need it to be an empty string for 
+	// resetting to initial after navigating away from password setting page.
+	userRecord.record.password = '';
 
 	if ( userRecord.hasEdits ) {
 		passwordStrong = isPasswordStrong(
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -40,9 +40,6 @@ export default function Password() {
 	const [ inputType, setInputType ] = useState( 'password' );
 	const [ hasAttemptedSave, setHasAttemptedSave ] = useState( false );
 	let passwordStrong = true; // Saved passwords have already passed the test.
-	// Originally, password wasn't set in record, but we need it to be an empty string for 
-	// resetting to initial after navigating away from password setting page.
-	userRecord.record.password = '';
 
 	if ( userRecord.hasEdits ) {
 		passwordStrong = isPasswordStrong(
--- a/settings/src/utilities.js
+++ b/settings/src/utilities.js
@@ -12,6 +12,11 @@ export function getUserRecord( userId ) {
 		userRecord.isSaving = useSelect( ( select ) => select( coreDataStore ).isSavingEntityRecord( 'root', 'user', userId ) );
 	}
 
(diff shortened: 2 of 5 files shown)


settings/src/components/password.js
```diff
@@ -40,6 +40,9 @@ export default function Password() {
 	const [ inputType, setInputType ] = useState( 'password' );
 	const [ hasAttemptedSave, setHasAttemptedSave ] = useState( false );
 	let passwordStrong = true; // Saved passwords have already passed the test.
+	// Originally, password wasn't set in record, but we need it to be an empty string for 
+	// resetting to initial after navigating away from password setting page.
+	userRecord.record.password = '';
```
StevenDufresne: Should this be here or in `utilities.js`: https://github.com/WordPress/wporg-two-factor/blob/31cc878ad267049b8f0ecda17a80a97bd7bc7aaf/settings/src/utilities.js#L7

StevenDufresne (changes requested):
Can you add a test for this?

settings/src/components/password.js
outdoor2kode: ~I think both would work, but I'm tempted to put it in `password.js`, as~
~1. the `record.password` value always needs to be assigned empty for the purpose of clearing the new password input here.
2. it's easier to locate it and do some tweaks once we decide to add an `old password` to the view - so that users have to verify the old password before changing to a new one.
3. as 2. said, if we're going to add an old password, it would be more appropriate to change it to userRecord.record.newPassword here.
4. there would never be a newPassword in the state `userRecord.record`. (there might be record.password in the future if we decide to add in the old password)~

**Update**

1. Upon further thought, I believe it's more reasonable to move all userRecord (state-related) processing into the `getUserRecord` function, allowing components to avoid dealing with state-manipulation stuff directly.
2. Tests added.

settings/src/utilities.js
```diff
@@ -12,6 +12,11 @@ export function getUserRecord( userId ) {
 		userRecord.isSaving = useSelect( ( select ) => select( coreDataStore ).isSavingEntityRecord( 'root', 'user', userId ) );
 	}
 
+	// Initialize the password as an empty string, necessary for resetting incomplete state when leaving the password setting page.
+	if ( userRecord.record && undefined === userRecord.record.password ) {
```
StevenDufresne, line 16: @ryelle @adamwoodnz Have we preferred optional chaining in our code bases?

StevenDufresne (approved):

StevenDufresne:
Thanks for the tests!

settings/src/utilities.js
outdoor2kode: I've tried optional chaining, but it didn't work well in this case.

When `useEntityRecord` is called for the first time, `userRecord.record` is undefined. This causes the code to enter the if condition directly and results in an error: `Cannot read properties of undefined (reading 'password')`

We could probably add something like the below to this [Line](https://github.com/WordPress/wporg-two-factor/pull/136/files/fa879b64b45171f2e3c5c9843bc218dc0c393109#diff-d2b2e322863e6dc961ea2ef9b6f6b6e2b5439e53ceb4144f5805b1dd61411b16R9) to fix it.
```
if ( ! userRecord.hasResolved ) {
  return;
}
```

PR #138, opened 2023-04-28
Fix eslint

### LINKED ISSUE #137: JS code quality with eslint
We depend on `@wordpress/scripts` and have the [package.json scripts in place for linting](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/package.json#L12), but we appear to be ignoring the eslint errors and problems, which can lead to subtle tricky bugs. In particular I've noticed we're ignoring some violations of the rules of hooks, eg.

Screen Shot 2023-04-28 at 10 29 50 AM

Running `yarn lint:js` reports the following:

Screen Shot 2023-04-28 at 10 36 33 AM

Adding an `.eslintrc.js` as described [here](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-eslint-plugin/#usage) is all that's required to have these surfaced in most editors.

We should autofix all those that can be, then fix any other critical issues, and adjust some rules or individually ignore things we don't consider worth fixing.

Once errors are addressed we should fail branch builds with errors.

adamwoodnz:
Fixes #137 

Adds the `@wordpress/eslint-plugin/recommended` settings which includes Prettier formatting. See the commits for the steps I've taken to fix the 396 problems originally identified.

After auto fixing eslint/prettier problems, and the trivial fixes, there are 8 errors remaining which are likely to cause bugs:

```
$ wp-scripts lint-js

/home/runner/work/wporg-two-factor/wporg-two-factor/settings/src/components/email-address.js
Error:   48:4  error  Unexpected alert        no-alert
Error:   48:4  error  'alert' is not defined  no-undef

/home/runner/work/wporg-two-factor/wporg-two-factor/settings/src/components/totp.js
Error:   144:23  error  React Hook "useCallback" is called conditionally. React Hooks must be called in the exact same order in every component render                                                                 react-hooks/rules-of-hooks
Error:   182:23  error  React Hook "useCallback" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?  react-hooks/rules-of-hooks

/home/runner/work/wporg-two-factor/wporg-two-factor/settings/src/script.js
Error:    77:3  error  'history' is not defined  no-undef
Error:   11[9](https://github.com/WordPress/wporg-two-factor/actions/runs/4825873555/jobs/8597161531?pr=138#step:5:10):4  error  'history' is not defined  no-undef

/home/runner/work/wporg-two-factor/wporg-two-factor/settings/src/utilities.js
Error:   [10](https://github.com/WordPress/wporg-two-factor/actions/runs/4825873555/jobs/8597161531?pr=138#step:5:11):21  error  React Hook "useEntityRecord" is called in function "getUserRecord" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"  react-hooks/rules-of-hooks
Error:   [14](https://github.com/WordPress/wporg-two-factor/actions/runs/4825873555/jobs/8597161531?pr=138#step:5:15):[25](https://github.com/WordPress/wporg-two-factor/actions/runs/4825873555/jobs/8597161531?pr=138#step:5:26)  error  React Hook "useSelect" is called in function "getUserRecord" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"        react-hooks/rules-of-hooks

✖ 8 problems (8 errors, 0 warnings)
```

## How to test

Regression test the app. The 8 non-trivial errors were related to saving email address and switching between QR code and manual totp setup methods.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,31 @@
+name: Static Analysis (Linting)
+
+# This workflow is triggered on pushes to trunk, and any PRs.
+on:
+  push:
+    branches: [fix/137-eslint]
```
adamwoodnz: Update this once we're ready to run on all branches

settings/src/components/email-address.js
```diff
@@ -18,23 +14,24 @@ import { GlobalContext } from '../script';
  * Render the Email setting.
  */
 export default function EmailAddress() {
-	const { userRecord }                                           = useContext( GlobalContext );
+	const { userRecord } = useContext( GlobalContext );
 	const { record, edit, save, editedRecord, hasEdits, isSaving } = userRecord;
-	const [ emailError, setEmailError ]                            = useState( '' );
-	const [ justChangedEmail, setJustChangedEmail ]                = useState( false );
+	const [ emailError, setEmailError ] = useState( '' );
+	const [ justChangedEmail, setJustChangedEmail ] = useState( false );
```
dd32, line 20: Is it possible for us to avoid these whitespace changes? 

I get that JS and PHP coding standards don't match, but aligning variable assignments makes for so much more readable code..

adamwoodnz: Sure I'll see if we can enforce that

adamwoodnz: It seems that aligning variable assignments isn't possible with Prettier, which is included by default in `@wordpress/eslint-plugin/recommended`. Closest issue I've found is https://github.com/prettier/prettier/issues/1622

Personally I'm used to this formatting, but I know it irks others. If we feel really strongly we could use Prettier to get things in the shape we want in this PR, then disable it in our eslint configuration.

```javascript
module.exports = {
	extends: 'plugin:@wordpress/eslint-plugin/recommended',

	rules: {
		...,
		'prettier/prettier': 0,
	},
};
```

Essentially the current state, but I'll revert these particular whitespace changes.

settings/src/edit.js
```diff
@@ -10,11 +10,13 @@ import { Placeholder } from '@wordpress/components';
...
 export default function Edit() {
 	return (
-		<div { ...useBlockProps() } >
+		<div { ...useBlockProps() }>
 			<Placeholder
 				instructions={ __( 'Not rendered in editor.', 'wporg' ) }
 				label={ __( 'Two-Factor Authentication', 'wporg' ) }
-			> </Placeholder>
+			>
+				{ ' ' }
```
StevenDufresne: This should be removed.

settings/src/utilities.js
```diff
@@ -3,14 +3,15 @@ import { store as coreDataStore, useEntityRecord } from '@wordpress/core-data';
...
+export function useGetUserRecord( userId ) {
+	const userRecord = useEntityRecord( 'root', 'user', userId );
 
-	// Polyfill in isSaving.
-	if ( undefined === userRecord.isSaving ) {
-		userRecord.isSaving = useSelect( ( select ) => select( coreDataStore ).isSavingEntityRecord( 'root', 'user', userId ) );
-	}
+	userRecord.isSaving = useSelect( ( select ) =>
+		select( coreDataStore ).isSavingEntityRecord( 'root', 'user', userId )
+	);
```
adamwoodnz, line 14: Calling hooks conditionally can have unpredictable results, and I'm not seeing any changes in behaviour on the frontend when calling this unconditionally. I believe keeping this value in sync with the store at all times is desirable. @dd32 @iandunn is there a reason this has been written this way? Is it intended to be a perf optimization?

settings/src/edit.js
adamwoodnz: Done

settings/src/components/email-address.js
```diff
@@ -47,44 +43,43 @@ export default function EmailAddress() {
 		try {
 			await edit( { pending_email: '' } );
 			await save();
-		} catch( error ) {
-			alert( error.message );
+		} catch ( error ) {
+			setEmailError( error.message );
```
adamwoodnz, line 47: I'm unsure whether this is an appropriate message to be displayed to a user or what it'll look like. Needs testing.

adamwoodnz: Tested this by returning a 404 and it appears it'll handle it appropriately

Screen Shot 2023-05-01 at 4 19 18 PM

settings/src/components/email-address.js
adamwoodnz: Discussed elsewhere: the team favours following Gutenberg standards and keeping Prettier enabled.

settings/.eslintrc.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+	extends: 'plugin:@wordpress/eslint-plugin/recommended',
+
+	rules: {
+		'jsdoc/require-param-type': 0,
```
adamwoodnz, line 5: This disables the requirement to have types declared on all function parameters. I've decided to disable for now as it would be a significant effort. We could add these at a later date to better document what's expected.

settings/.eslintrc.js
```diff
@@ -0,0 +1,14 @@
...
+		'prettier/prettier': [
+			'error',
+			{
+				...require( '@wordpress/prettier-config' ),
+				printWidth: 100,
```
adamwoodnz, line 10: Default is 80, which resulted in some of the longer strings having `{' '}` inserted when React components are inline within the text. 100 seems like a reasonable middle ground where this is rare.

settings/src/utilities.js
iandunn: It looks like it was added in 2925a2f66597b3bb7f07e4a54c1c7d85e9032285. I'm not sure why, but if you're not seeing any problems then it seems safe to go ahead and change it. We can always dig deeper if we notice a problem in the future 👍🏻

dd32 (approved):
While I might not agree 100% with some of the formats chosen by the JS formatter, I agree that maintaining style with Gutenberg and other WP JS projects is worth the tradeoffs.

Having some kind of standard is better than none.

.github/workflows/lint.yml
adamwoodnz: Done

Issue #131, opened 2023-04-24
Inline password strength indicator.


StevenDufresne:
Related to #129, #118 

We currently have an error message that is somewhat disruptive when creating a new password. Let's consider improving the interface with a better inline approach.

The inline approach should:
- Indicate when a password is strong/weak
- Inform the user what should be added to make password stronger.

We do these things already, but the indicator is a bit disconnected from the input and has a impactful change on layout in its current state.

### Current:

| Success | Error |
| --- | --- |
| <img width="724" alt="Screenshot 2023-04-24 at 11 07 18 AM" src="https://user-images.githubusercontent.com/1657336/233883618-df9dc3ea-6abd-453c-8f67-db00d73e2858.png"> | <img width="725" alt="Screenshot 2023-04-24 at 11 07 10 AM" src="https://user-images.githubusercontent.com/1657336/233883629-e59f14fc-6f35-49b0-8922-8ac73b227b6e.png"> |

PR #141, opened 2023-04-28
WebAuthn: Link to wp-admin until custom UI is built.

iandunn:
See #114
Based on #95

This adds a stub for the custom WebAuthn UI, so that users will be aware it is an option. It's behind a feature flag until all the other PRs associated with #114 are done. At that point, it can be enabled for everyone.

This may need some design/copy tweaks, but it's temporary, so IMO it's best to not spend too much time on it.

Screenshot 2023-04-28 at 12 43 42 PM

Screenshot 2023-04-28 at 12 43 57 PM

### DIFF
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -19,6 +19,7 @@ export default function AccountStatus() {
 	const { record }        = userRecord;
 	const emailStatus       = record.pending_email ? 'pending' : 'ok';
 	const totpStatus        = record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' ) ? 'enabled' : 'disabled';
+	const webAuthnStatus    = record[ '2fa_available_providers' ].includes( 'Two_Factor_WebAuthn' ) ? 'enabled' : 'disabled';
 	const backupCodesStatus = record[ '2fa_available_providers' ].includes( 'Two_Factor_Backup_Codes' ) ? 'enabled' : 'disabled';
 
 	return (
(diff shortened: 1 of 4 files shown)


jasmussen:
CC: @fcoveram as it may relate to some of the "password strength" mockups you did.

iandunn:
Did you mean to leave that comment on #131 ? If not, I'm confused about how it relates to this PR. Do you mind clarifying?

jasmussen:
Sorry, I did not mean to be intentionally obtuse. I mean that the iconography (check vs. X), the use of color and notices, and the indicators for whether something is complete or incomplete (in this case, steps in the stepper) could/should arguably share DNA with the same iconography and indication we use for password strength, if not now then at least at some point in the future.

Here are a couple of quick snaps from Francisco's work, which relies more on the high contrast which gray icons and text can give, compared to stoplight colors:

Screenshot 2023-05-01 at 18 13 36

Screenshot 2023-05-01 at 18 13 57
Screenshot 2023-05-01 at 18 14 02

iandunn:
Ah, that makes sense, thanks! Do you mind if I create a new issue for that, since it applies to all of the UI, rather than something specific to this PR?

StevenDufresne:
Thanks for the ticket. I'm thinking we can omit this from the MVP launch in favor of releasing faster and managing expectations. As soon as we release this view, we'll be "on the clock" to switch it out. Instead, maybe we could mention the /wp-admin path for WebAuthn in the beta/alpha launch post and iterate with more data. Thoughts?

fcoveram:
@StevenDufresne I'm fine with delaying the style tweak and mentioning it for the upcoming changes

Thanks @iandunn. Please ping me in the ticket you will open to share a design idea.

StevenDufresne:
> @StevenDufresne I'm fine with delaying the style tweak and mentioning it for the upcoming changes
> 
> Thanks @iandunn. Please ping me in the ticket you will open to share a design idea.

Just for clarity, I was talking about the list item "Two Factor Security Key" in the description. I don't think that is necessary for launch. 

The second thread revolves around style treatment and @iandunn's has offered to open a new ticket regarding that. I am also in agreement that we could also iterate on that. 👍

iandunn:
Thanks! I created #145 for the that @fcoveram.  The changes don't necessarily have to be _delayed_, it's just a lot easier to manage when there are separate issues for unrelated things. In some cases that can actually make it faster, because someone else on the team can pick it up and work on it in parallel.

> Instead, maybe we could mention the /wp-admin path for WebAuthn in the beta/alpha launch post and iterate with more data

That sounds good, I'll go ahead and close this 👍🏻

PR #142, opened 2023-04-28
Tests: Require `wporg-mu-plugins` to run tests.

### LINKED ISSUE #140: PHPUnit tests failing
```
There were 2 failures:

1) Test_WPorg_Two_Factor::test_set_primary_provider_for_user
Failed asserting that two strings are not identical.

wporg-two-factor/tests/test-wporg-two-factor.php:250

2) Test_WPorg_Two_Factor::test_totp_key_upgraded_to_encrypted
Failed asserting that two strings are not identical.

wporg-two-factor/tests/test-wporg-two-factor.php:290

FAILURES!
Tests: 13, Assertions: 48, Failures: 2.
Script phpunit --no-coverage handling the test event returned with error code 1
```

It looks like those were introduced in aae7ec97a9a8ee22342476f2e0bc78ac3ff4aa3c, cc @dd32. Let me know if the environment is still making it difficult to get the tests setup. I've got it working locally and am happy to take a look.

iandunn:
Fixes #140

This requires `wporg-mu-plugins` to be installed because, without it, tests will fail incorrectly. Instructions for setting it up are in `README.md`, so I think it's safe to assume that it should exist. 

If folks think it might trip people up, though, we could output an error message instead of letting it fatal. That could point people in the right direction if they're not sure why it's failing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ WordPress.org-specific customizations for the Two Factor plugin
 		return in_array( $user->user_login, $GLOBALS['supes'], true );
 	}
 	```
-1. Install and build the `wporg-mu-plugins` repository.
+1. `git clone https://github.com/WordPress/wporg-mu-plugins.git --branch build wp-content/mu-plugins/pub`
 1. Add this code to your `wp-content/mu-plugins/0-sandbox.php`:
 	```php
 	require_once WPMU_PLUGIN_DIR. '/wporg-mu-plugins/mu-plugins/loader.php';
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ WordPress.org-specific customizations for the Two Factor plugin
 		return in_array( $user->user_login, $GLOBALS['supes'], true );
 	}
 	```
-1. `git clone https://github.com/WordPress/wporg-mu-plugins.git --branch build wp-content/mu-plugins/pub`
 1. Add this code to your `wp-content/mu-plugins/0-sandbox.php`:
 	```php
 	require_once WPMU_PLUGIN_DIR. '/wporg-mu-plugins/mu-plugins/loader.php';
--- a/composer.json
+++ b/composer.json
@@ -10,16 +10,29 @@
 			"php": "7.4"
 		},
 		"allow-plugins": {
-			"dealerdirect/phpcodesniffer-composer-installer": true
+			"dealerdirect/phpcodesniffer-composer-installer": true,
+			"composer/installers": true
 		}
 	},
-	"require": {},
+	"extra": {
+		"installer-paths": {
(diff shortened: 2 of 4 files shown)


iandunn:
Converting back to draft while I look into action failures. Looks like a difference between the wp-env environment and mine.

iandunn:
Ok, I think de75c8d is a better approach. I also needed to move my `wporg-mu-plugins` checkout to the new path in the readme, so that it matches `wp-env`.

I noticed some other things that could be improved about the setup to make local and `wp-env` environments match, and logged those in #143.

dd32:
Personally, I don't want to support multiple ways of setting up local environments. wp-env has it's problems, but it's intentionally biased and we're reliant upon that. I'm in favour of removing the other steps from the readme and putting it into an errata or other document instead as an `alternative setup options`.

Looking at this diff, there's two things that jump out at me:
 1. The additional `require` shouldn't be required, as it's an mu-plugin it should be loaded as part of the WP install that the wp-phpunit requires
 2. again.. The existing `require`s of the plugin shouldn't be required, as the plugin should be loaded on the WP install.

But that being said, this has zero impact upon the existing `wp-env` usage, so I've got nothing against adding it, only against trying to support multiple environment types.

StevenDufresne:
I would also vote we treat this the same way we do with the other themes and install it via composer. Refs: [wporg-main-2022:composer.json](https://github.com/WordPress/wporg-main-2022/blob/499067319e790420546b021f0e3c72622a0e45f3/composer.json#L102), [wporg-main-2022:0-sandbox.php](https://github.com/WordPress/wporg-main-2022/blob/499067319e790420546b021f0e3c72622a0e45f3/env/0-sandbox.php#L9)

iandunn:
I'm In favor of moving those steps into an `alternative setup` section, I've added that to #143 👍🏻 

I think installing `wporg-mu-plugins` via `composer.json` is a good idea too, I'll add that to this PR 👍🏻 

I also like the idea of a single environment in theory, and I've tried switching to wp-env and other Docker-based envs several times in the past. Each time, though, I've found that it doesn't work well for me personally, and it's just simpler to [run nginx locally](https://noisysocks.com/2021/11/12/set-up-a-wordpress-development-environment-using-homebrew-on-macos/).

I'm happy to work around it though if you all would prefer to keep things as they are.

iandunn:
I added the Composer checkout in 5a36207, but [GitHub Actions have a partial outage right now](https://www.githubstatus.com/incidents/pr3498h3qkfy). The tests pass in my local env, but I'm going to wait until that's resolved to make sure they pass here too.

I'm planning to merge if they are, but let me know if either of you would like to discuss something different.

PR #144, opened 2023-05-02
Enable-2FA notice shown when already setup

dd32:
remove_capabilities_until_2fa_enabled() doesn't verify that the accounts capabilities being filtered is the current user, as a result, running `user_can( $other_user, 'admin_only_cap' )` will cause this plugin to add the enable-2fa notice.

- [ ] Unit tests

### DIFF
--- a/wporg-two-factor.php
+++ b/wporg-two-factor.php
@@ -114,7 +114,7 @@ function remove_super_admins_until_2fa_enabled() : void {
  * perform privileged actions on the front end, via the REST API, etc.
  */
 function remove_capabilities_until_2fa_enabled( array $allcaps, array $caps, array $args, WP_User $user ) : array {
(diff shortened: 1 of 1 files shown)


dd32:
I'm not convinced this PR is correct anymore.

I do suspect the notice is still being added incorrectly, and should be gated like so:
```
		if ( get_current_user_id() === $user->ID ) {
			add_action( 'admin_notices', __NAMESPACE__ . '\render_2fa_admin_notice' );
			add_filter( 'wporg_global_header_alert_markup', __NAMESPACE__ . '\get_enable_2fa_notice' );
		}
```

I can duplicate the notice appearing incorrectly still when viewing https://wordpress.org/support/plugin/test-plugin-3/reviews/ 

I haven't been able to create a unit test that shows the behaviour however..

dd32:
This was merged in #319

PR #147, opened 2023-05-04
Prompt for revalidation of 2FA details.

### LINKED ISSUE #43: Reauth 2nd factor to change 2FA settings
There's [an upstream issue for this](https://github.com/WordPress/two-factor/issues/484), but we'll probably need to do our own since we have the custom UI.

### LINKED ISSUE #115: Reauth 2FA UI
Ref: https://github.com/WordPress/wporg-two-factor/issues/43#issuecomment-1496978570

## Acceptance Criteria
Prompt user when it's been > xx mins since the last 2FA challenge before displaying UI.
* Question: Done by the back-end side? If not, we might need to `Store the timestamp of the last 2FA challenge: When the user successfully completes a 2FA challenge, store the timestamp (e.g., as a Unix timestamp), such as storing in a cookie, local storage or Recat state.`

dd32:
Depends upon https://github.com/WordPress/two-factor/pull/529

Fixes #43 #115

Require revalidation of the user via two-factor if accessing two-factor settings or backup codes, and the `userRecord.record[ '2fa_revalidation' ].expires_at` time has passed.

TODO:
 - [x] Wording
 - ~[] Translations~
 - [x] Merge https://github.com/WordPress/two-factor/pull/529


dd32:
The current UI only enforces 2FA revalidation for TOTP and Backup codes, as Email and Password don't need it.

The current code doesn't block a user attempting to setup TOTP or generating backup codes, as while the user could generate a TOTP key and URI, they're unable to save it via the Two Factor plugin (or be able to access the Backup Codes API)

Due to `sameorigin` policies, using an inline iframe isn't an option, at present, but potentially could be in the future if we wanted to go down that route.

dd32:
> Due to sameorigin policies, using an inline iframe isn't an option, at present, but potentially could be in the future if we wanted to go down that route.

I realised with some [minor changes to the upstream patch](https://github.com/WordPress/two-factor/pull/529/commits/2d9e61bd868a919a3b7530c0fb41b5cd41f510b9), [and to WordPress.org's SSO](https://github.com/WordPress/wordpress.org/commit/97f69aecd9dbe36c969e38278cee10ac5008c36f), I could make this work. 

I have **not** tested this with a Security key, but due to the Origin being the same, I believe this should work based on some testing using other sites.

https://user-images.githubusercontent.com/767313/236746414-b6b9cd1a-ea56-4c08-8f69-0b2b10ecd5bd.mov

The delay in the above recording is while it's doing a `refreshUser` API call, I suspect this can be made much nicer somehow, perhaps just by having the `postMessage` pass through the new expiry data and have it refresh it in the background, or have it assume that it's okay.

dd32:
With the addition of [3ca07e1](https://github.com/WordPress/wporg-two-factor/pull/147/commits/3ca07e1c1194d7d016a0d58e662184588909e6c3) the UI feels much 'faster' and a far better UX.

There's obviously going to be some edge-cases in that, but it appears as if it works well enough.

I haven't tested this PR at all with setting up TOTP, I know it 'works' with the Two-Factor plugin, but I haven't tested our UI with it.

https://user-images.githubusercontent.com/767313/236967916-ace7ed22-c5fd-4802-b658-4411ae5629b2.mov

dd32:
Frustratingly I just realised that this branch appears to be based on an old `trunk`.. I'll merge/rebase this today. _edit: turns out not to have been a problem.._

However, I can confirm that the iframe approach works with WebAuthN as in #153

settings/src/components/revalidate-modal.js
```diff
@@ -0,0 +1,80 @@
+/**
+ * WordPress dependencies
+ */
+import { useContext, useEffect, useRef, useState } from '@wordpress/element';
+import { GlobalContext } from '../script';
+import { Button, Modal, __experimentalHStack as HStack } from '@wordpress/components';
```
adamwoodnz: I saw talk of this being marked stable as it's used 'all over the place' now. Can probably eslint ignore this particular failure.

dd32: That's what I was thinking, it appears to be one of the "stable" unstable API's now, and why I used it in the first place.

I'm not sure how to mark it as ignored though yet.

adamwoodnz: Changed my mind as it looks like [things are still unstable](https://github.com/WordPress/gutenberg/issues/43085)

adamwoodnz:
I'm seeing this warning @dd32, not sure whether it's a concern

Screen Shot 2023-05-10 at 11 40 49 AM

adamwoodnz:
Latest UI:
- Single modal flow and copy matching [design](https://github.com/WordPress/wporg-two-factor/issues/115#issuecomment-1521317766), but with restyled login page 2FA elements
- Account Status is displayed behind modal

2FA reauth

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8000_users_adamwood_edit_account__screen=account-status(Desktop) | localhost_8000_users_adamwood_edit_account__screen=account-status(iPad) | localhost_8000_users_adamwood_edit_account__screen=account-status(Samsung Galaxy S20 Ultra) |

adamwoodnz:
@dd32 there is a translation task in the description but I don't see any other translations in the JS app. What are we expecting here?

dd32:
>  there is a translation task in the description but I don't see any other translations in the JS app.

Yeah looks like we ditched translations, we can skip that part then for now :)

dd32:
> I'm seeing this warning dd32, not sure whether it's a concern

@adamwoodnz Thanks! I'm not sure how I missed that 🤔; I fixed the variable reference, it was a copy-paste error.

> Latest UI

Looks good to me! Thanks!

I think we should do some extra work on the 2FA prompts displayed during login too, but that's not needed as part of this, and can wait for the next Iteration.

adamwoodnz:
> Yeah looks like we ditched translations, we can skip that part then for now :)

Cool, guess we're just dependent on the two-factor PR being merged then. Let me know when that happens, or feel free to merge this yourself of course :)

StevenDufresne:
I wasn't able to get my environment configured to test this out, but based on what I see, I think this is good solution for the MVP since the grace time should mean that not many users go through this flow (so it's not super critical), but their accounts are protected. 

The modal does feel like it has a long chin (a lot of whitespace at the bottom), but I understand it's related to embedding a variable width iframe. 

As part of merging this ticket, do ya'll mind opening a new ticket with information about implementing a more ideal case while it's fresh in your mind?

dd32:
> As part of merging this ticket, do ya'll mind opening a new ticket with information about implementing a more ideal case while it's fresh in your mind?

I'm not sure what a better solution is really, without too much code duplication.

The only thing I think that could realistically be done here is to improve the actual Prompt screen to not have as much screen reflow (ie. not insert an error div, but rather have it take up some space that is currently `visibility: hidden`), and have the postMessage communicate the ideal sizes for the embedded iframe.

PR #154, opened 2023-05-08
Add a Paste handler for the TOTP interface.

### LINKED ISSUE #151: Can't paste TOTP codes
This is a common workflow for me with TOTP on mobile:

1. Open TOTP app
2. Longpress to copy current TOTP code
3. Switch to app that requested TOTP code
4. Tap the input field, then tap the clipboard-autocomplete button, so that the code is pasted in for me

One downside of #100 is that workflow is no longer possible, and users now need to manually enter each digit. On mobile that might involve switching back and forth between apps if the user forgets part of the code.

Maybe we can add something that watches for a paste event on the 1st input field, and the fills all of the fields? I worry that folks would still be discouraged from attempting to paste, though, because having separate input fields implies that you can't paste into all of them at once.

Maybe there's a way to return to a single input field, but style it in a way that matches the design? If not, I think it may be worth reconsidering having separate input fields.

dd32:
Fixes #151 

https://user-images.githubusercontent.com/767313/236757157-e43c6a12-5109-4152-b581-a371f020086d.mov

The addition of the `onComplete()` is to trigger the Enable button to enable. I've just noticed as reviewing the video above that the clear button leaves Enable button enabled..

### DIFF
--- a/settings/src/components/auto-tabbing-input.js
+++ b/settings/src/components/auto-tabbing-input.js
@@ -9,27 +9,13 @@ import { useCallback } from '@wordpress/element';
 import NumericControl from './numeric-control';
 
 const AutoTabbingInput = ( props ) => {
-	const { inputs, setInputs, onComplete, error } = props;
+	const { inputs, setInputs, error } = props;
 
 	const handleChange = useCallback( ( value, event, index, inputRef ) => {
 		setInputs( ( prevInputs ) => {
 			const newInputs = [ ...prevInputs ];
 
-			// Clean input
-			if ( value.trim() === '' ) {
-				event.target.value = '';
-				value = '';
-			}
-
-			newInputs[ index ] = value;
-
-			// Check if all inputs are filled
-			const allFilled = newInputs.every( ( input ) => '' !== input );
-			if ( allFilled && onComplete ) {
-				onComplete( true );
-			} else {
-				onComplete( false );
-			}
+			newInputs[ index ] = value.trim() === '' ? '' : value;
 
 			return newInputs;
 		} );
@@ -51,7 +37,6 @@ const AutoTabbingInput = ( props ) => {
 		if ( inputs.length === paste.length ) {
 			event.preventDefault();
 			setInputs( paste.split( '' ) );
-			onComplete( true );
 		}
 	}, [] );
 
--- a/settings/src/components/auto-tabbing-input.js
+++ b/settings/src/components/auto-tabbing-input.js
@@ -46,17 +46,20 @@ const AutoTabbingInput = ( props ) => {
 	}, [] );
 
 	const handlePaste = useCallback( ( event ) => {
-		const paste = event.clipboardData.getData('Text').replace( /[^0-9]/g, '' )
+		const paste = event.clipboardData.getData( 'Text' ).replace( /[^0-9]/g, '' );
 
-		if ( inputs.length == paste.length ) {
+		if ( inputs.length === paste.length ) {
 			event.preventDefault();
-			setInputs( paste.split('') );
+			setInputs( paste.split( '' ) );
 			onComplete( true );
 		}
 	}, [] );
 
 	return (
-		<div className={ 'wporg-2fa__auto-tabbing-input' + ( error ? ' is-error' : '' ) } onPaste={ handlePaste }>
+		<div
+			className={ 'wporg-2fa__auto-tabbing-input' + ( error ? ' is-error' : '' ) }
+			onPaste={ handlePaste }
+		>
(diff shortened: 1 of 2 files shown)


settings/src/components/auto-tabbing-input.js
```diff
@@ -45,8 +45,21 @@ const AutoTabbingInput = ( props ) => {
 		}
 	}, [] );
 
+	const handlePaste = useCallback( ( event ) => {
+		const paste = event.clipboardData.getData( 'Text' ).replace( /[^0-9]/g, '' );
+
+		if ( inputs.length === paste.length ) {
```
adamwoodnz: I wonder if it makes sense to show a message if the length doesn't match, eg. 'Please enter a six digit code'. I know there is text above instructing this, but without it I think it's hard to understand what's going wrong, say in the case where they have made a mistake copying.

We already have a notice component there which could display the message.

adamwoodnz (commented):
Works for me. I tested longer and shorter strings and I felt like the validation should result in a message for the user.

settings/src/components/auto-tabbing-input.js
adamwoodnz: It's also possible to paste letters in, we might need validation for that too, unless they're valid

Screen Shot 2023-05-09 at 11 19 04 AM

settings/src/components/totp.js
```diff
@@ -225,7 +225,10 @@ function SetupForm( { handleEnable, qrCodeUrl, secretKey, inputs, setInputs, err
 	}, [ error, inputs ] );
 
 	const handleComplete = useCallback( ( isComplete ) => setIsInputComplete( isComplete ), [] );
```
adamwoodnz: Unrelated, but from my understanding there is no benefit to using `useCallback` with a `useState` setter function, as it is stable between renders. We can pass `setIsInputComplete` to the child component directly.

adamwoodnz:
Happy to work on this if it would take a load off you @dd32

settings/src/components/auto-tabbing-input.js
dd32: > I wonder if it makes sense to show a message if the length doesn't match, eg. 'Please enter a six digit code'. 

I thought about whether it should do something in that scenario, but decided against it and just went for the whole "If you paste what we're expecting, process it, else let the rest of the UI code handle it"

> It's also possible to paste letters in, we might need validation for that too, unless they're valid

It's also possible to type letters in though, so I didn't really see that as much of an issue.. That might vary between browsers. I've only tested with Chrome.

settings/src/components/totp.js
dd32: I don't know React very well :) but that sounds good to me.

dd32:
> Happy to work on this if it would take a load off you dd32

@adamwoodnz In general, please feel free to pick up anything I PR (or push extra commits to, to fix something) especially if it's UI/JS related. I'll _normally_ leave things as a Draft or explicitly mention it's a `WIP` if I'm still working on it :)

settings/src/components/auto-tabbing-input.js
adamwoodnz: > It's also possible to type letters in though

Turns out this is validated on submit, so yes I think we can leave this alone as a rare edge case.

I have added a length validation error message though.

settings/src/components/auto-tabbing-input.js
```diff
@@ -9,27 +9,13 @@ import { useCallback } from '@wordpress/element';
...
-			newInputs[ index ] = value;
-
-			// Check if all inputs are filled
-			const allFilled = newInputs.every( ( input ) => '' !== input );
-			if ( allFilled && onComplete ) {
-				onComplete( true );
-			} else {
-				onComplete( false );
-			}
+			newInputs[ index ] = value.trim() === '' ? '' : value;
```
adamwoodnz, line 18: Rather than modifying the value of the DOM element here, we can simply update the `inputs`, and the data will flow down to the DOM.

The `complete` state can also be derived from the `inputs` values, rather than checking on every change to an input, see the check for every input having a value [here](https://github.com/WordPress/wporg-two-factor/pull/154/files#diff-9b176e953d93ade24d65eef39d9a5ebb31bec9e284b349cf8ff313f6e9869f8fR235)

adamwoodnz:
Desktop UI with:

1. Type incorrect totp, submit
2. Type change but still incorrect, submit
3. Paste totp too short over existing (no inputs change, error shown)
4. Clear and paste totp too short (no inputs change, error shown)
5. Paste totp correct length, submit
6. Final correct totp

paste totp

iandunn (approved):
LGTM 👍🏻

PR #157, opened 2023-05-11
Disable Backup Codes UI until primary provider enabled

### LINKED ISSUE #47: Disable backup codes UI until primary provider enabled

iandunn:
Fixes #47

This disables the Backup Codes UI when the user hasn't enabled TOTP yet. In the future, WebAuthn can be included so that any primary provider is acceptable, rather than TOTP specifically.

Related #158  -- The UI will show Backup Codes as enabled when TOTP is disabled, but the root cause for that is outside the scope of this PR

### DIFF
--- a/settings/.eslintrc.js
+++ b/settings/.eslintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+	root: true,
 	extends: 'plugin:@wordpress/eslint-plugin/recommended',
 
 	rules: {
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -18,14 +18,20 @@ export default function AccountStatus() {
 	const { userRecord } = useContext( GlobalContext );
 	const { record } = userRecord;
 	const emailStatus = record.pending_email ? 'pending' : 'ok';
-	const totpStatus = record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' )
-		? 'enabled'
-		: 'disabled';
-	const backupCodesStatus = record[ '2fa_available_providers' ].includes(
-		'Two_Factor_Backup_Codes'
-	)
-		? 'enabled'
-		: 'disabled';
+	const totpEnabled = record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' );
+	const webAuthnEnabled = record[ '2fa_available_providers' ].includes(
+		'TwoFactor_Provider_WebAuthn'
+	);
+	const primaryProviderEnabled = totpEnabled || webAuthnEnabled;
+	const backupCodesEnabled =
+		record[ '2fa_available_providers' ].includes( 'Two_Factor_Backup_Codes' );
+
+	const backupBodyText =
+		! backupCodesEnabled && ! primaryProviderEnabled
+			? 'Please enable Two-Factor Authentication before enabling backup codes.'
+			: `You have
+				${ backupCodesEnabled ? '' : 'not' }
+				verified your backup codes for two-factor authentication.`;
 
 	return (
 		<>
@@ -49,10 +55,10 @@ export default function AccountStatus() {
 
 			<SettingStatusCard
 				screen="totp"
-				status={ totpStatus }
+				status={ totpEnabled }
 				headerText="Two-Factor Authentication"
 				bodyText={
-					'enabled' === totpStatus
+					totpEnabled
 						? /* @todo update this when hardware tokens become an additional option. */
 						  'You have two-factor authentication enabled using an app.'
 						: 'You do not have two-factor authentication enabled.'
@@ -61,11 +67,10 @@ export default function AccountStatus() {
 
 			<SettingStatusCard
 				screen="backup-codes"
-				status={ backupCodesStatus }
+				status={ backupCodesEnabled }
 				headerText="Two-Factor Backup Codes"
-				bodyText={ `You have ${
-					'enabled' === backupCodesStatus ? '' : 'not'
-				} verified your backup codes for two-factor authentication.` }
+				bodyText={ backupBodyText }
+				disabled={ ! backupCodesEnabled }
 			/>
 		</>
 	);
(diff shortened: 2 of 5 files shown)


settings/src/components/account-status.js
```diff
@@ -79,27 +86,31 @@ export default function AccountStatus() {
...
-						<Icon
-							icon={ chevronRight }
-							size={ 26 }
-							className="wporg-2fa__status-card-open"
-						/>
-					</CardBody>
-				}
-			/>
+		<Card className={ classes }>
+			{ disabled && cardContent }
```
StevenDufresne: I'm pretty sure you could do this with a ternary:
```
{ disabled ? cardContent : <ScreenLink /> }
```

settings/src/components/account-status.js
```diff
@@ -79,27 +86,31 @@ export default function AccountStatus() {
...
+			<StatusIcon status={ status } />
+			<h3 aria-label={ 'Click to enter the ' + headerText + ' setting page.' }>
+				{ headerText }
+			</h3>
+			<p>{ bodyText }</p>
+			<Icon icon={ chevronRight } size={ 26 } className="wporg-2fa__status-card-open" />
+		</CardBody>
+	);
+
+	let classes = 'wporg-2fa__status-card wporg-2fa__status-card-' + screen;
```
StevenDufresne, line 101: I'm not sure what our standards are in regards to concatenation, but maybe a template literal is a bit cleaner? 

Not a strong opinion:
```
`wporg-2fa__status-card wporg-2fa__status-card-${screen} `;
```

settings/src/components/account-status.js
```diff
@@ -21,12 +21,20 @@ export default function AccountStatus() {
 	const totpStatus = record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' )
 		? 'enabled'
 		: 'disabled';
+	const primaryProviderEnabled = 'enabled' === totpStatus; // @todo Update when webauthn UI built.
```
StevenDufresne: Since `totpStatus` is passed into `<SettingStatusCard>` as a string but used in other cases as a boolean, doesn't it make sense to change that logic? 

```
const primaryProviderEnabled = record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' );

....

<SettingStatusCard
  screen="totp"
  status={ primaryProviderEnabled ? 'enabled' : 'disabled' }
  headerText="Two-Factor Authentication"
  bodyText={
	  primaryProviderEnabled
		  ? /* @todo update this when hardware tokens become an additional option. */
		    'You have two-factor authentication enabled using an app.'
		  : 'You do not have two-factor authentication enabled.'
  }
/>

```

settings/.eslintrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+	root: true,
```
StevenDufresne, line 2: What's this change for?

adamwoodnz (commented):
I've previously had 2FA and backup codes enabled, but now I've disabled 2FA, should I still be able to access backup codes? I expected the logic to be based on 2FA being enabled

Screen Shot 2023-05-12 at 2 45 15 PM

settings/src/components/account-status.js
iandunn: That does work, but IMO the readability is lower:

`{ disabled ? cardContent : <ScreenLink screen={ screen } anchorText={ cardContent } /> }`

I feel like there's too much going on in that line to skim it quickly, I have to slow down to pull the different sections apart. I think the form in [9973b5f](https://github.com/WordPress/wporg-two-factor/pull/157/commits/9973b5f7f07bd6e3e7e1916db0c67a32de28cf37) has that problem to a lesser extent, but `prettier` wanted it that way, instead of allowing the attributes to have some breathe room on their own line:

```js
<Card className={ classes }>
	{ disabled && cardContent }

	{ ! disabled &&
		 <ScreenLink 
		 	screen={ screen }
			anchorText={ cardContent }
		/>
	}
</Card>
```

settings/src/components/account-status.js
iandunn: That feels a bit less readable to me, since there's extra symbols cluttering it, and they all run together for me. `prettier` adds spacing around `screen` though, which helps a fair amount:

```js
`wporg-2fa__status-card wporg-2fa__status-card-${ screen } `;
```

The Core standards favor [liberal amounts of spacing](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#spacing). 

I don't feel strongly, but am curious if there's a tangible benefit to template literals, or if it's just personal preference?

settings/.eslintrc.js
iandunn: It's to make sure eslint chooses this repo's file, rather than a fallback one higher up the directory structure. The commit message (b107bcfda6a186093b162702d453942c3faa714c) has a more detailed explanation.

iandunn:
>  now I've disabled 2FA, should I still be able to access backup codes?

No, but that felt outside the scope of this PR. I opened #158 for it.

settings/src/components/account-status.js
iandunn: 🤔 I don't think so, the `totp` status card needs to use `totpStatus` specifically, because if WebAuthn is enabled and TOTP is disabled, then `primaryProviderEnabled` would be `true` and `totpStatus` would be`'disabled'`. It looks like the code above would show the wrong status in that situation.

I added 747921c82a9eba07bdfb346cf9b6dc1273cd4fd8 to clarify how I imagine this will be used once we add WebAuthn.

Does that make sense, or did I misunderstand something?

StevenDufresne: Right, I understand how my code sample doesn't work. The main point is that it may be better to assign the status as a boolean since we are making checks based on it's string as opposed to a more predictable boolean.

```
const totpEnabled = record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' );
const webAuthnEnabled= record[ '2fa_available_providers' ].includes( 'TwoFactor_Provider_WebAuthn');
const primaryProviderEnabled = totpEnabled || webAuthnEnabled;

...

<SettingStatusCard
  screen="totp"
  status={ totpEnabled ? 'enabled' : 'disabled' }
  headerText="Two-Factor Authentication"
  bodyText={
	  totpEnabled
		  ? /* @todo update this when hardware tokens become an additional option. */
		    'You have two-factor authentication enabled using an app.'
		  : 'You do not have two-factor authentication enabled.'
  }
/>


```

settings/src/components/account-status.js
StevenDufresne: There are pros and cons for both techniques.... Personal preference :)

settings/src/components/account-status.js
StevenDufresne: It's pretty common to use a ternary for these states in react projects. You'll find it used often in Gutenberg and Gutenberg documentation ([example](https://github.com/WordPress/gutenberg/blob/trunk/docs/getting-started/create-block/author-experience.md)).

adamwoodnz: Yeah I agree this is a very common React pattern

settings/src/components/account-status.js
iandunn: That WFM, changed in 4ccb203 👍🏻

settings/src/components/account-status.js
iandunn: Yeah, I don't doubt it's common in general; this specific case just feels harder to read to me when doing it that way, and there isn't anything in [the Gutenberg guidelines](https://github.com/WordPress/gutenberg/blob/f8d74a6230b81a7980f492ba5c1a38cc23f28315/docs/contributors/code/coding-guidelines.md) or `prettier` about it. I went ahead and did it in becdb2c so we can get this out the door though.

---- 

More broadly, I feel like these types of discussions add a disproportionate amount of delay and debate to PRs, compared to the potential benefit of a minor style tweak. What do y'all think about an approach where we [agree on standards as a team once, and then let the tooling do the work of ensuring conformance](https://www.mattlayman.com/blog/2017/no-nitpicking-code-reviews/)? We could add a custom rule to the Gutenberg standards in a different PR if the team wants to.

Issue #127, opened 2023-04-21
Fix "Two-Factor Authentication" view console errors


StevenDufresne:
## Errors

```
react-dom.js?ver=18:1 Warning: React does not recognize the `setInputs` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `setinputs` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
...
react-dom.js?ver=18:1 Warning: Unknown event handler property `onComplete`. It will be ignored.
   ```

iandunn:
#152 is a duplicate of this, but I missed that when opening it. I'm gonna leave that one open since it has more info and is tied to the PR.

PR #159, opened 2023-05-11
Replace props spreading with explicit attributes to avoid warnings

### LINKED ISSUE #152: `setInputs / onComplete` gets passed to `<input>` element
I get a couple warnings locally because functions eventually end up being passed to a raw `<input>` element via `props` spreading.

> Warning: React does not recognize the `setInputs` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `setinputs` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

> Warning: Unknown event handler property `onComplete`. It will be ignored.

https://github.com/WordPress/wporg-two-factor/blob/e2bf00feeb90af48cab568ad1b9f5bc3c82bd632/settings/src/components/totp.js#L212

https://github.com/WordPress/wporg-two-factor/blob/e2bf00feeb90af48cab568ad1b9f5bc3c82bd632/settings/src/components/auto-tabbing-input.js#L52C1-L53

https://github.com/WordPress/wporg-two-factor/blob/e2bf00feeb90af48cab568ad1b9f5bc3c82bd632/settings/src/components/numeric-control.js#L35C1-L36

I don't think it causes any problems for the user, but it'd be nice to fix so that the dev console is less cluttered.

iandunn:
Fixes #152

This adds the necessary parameters explicitly, instead of spreading `props`.

Related #127

### DIFF
--- a/settings/src/components/numeric-control.js
+++ b/settings/src/components/numeric-control.js
@@ -18,7 +18,8 @@ import { useRef, useCallback } from '@wordpress/element';
  * @see https://stackoverflow.com/a/66759105/450127
  */
 export default function NumericControl( props ) {
-	const { autoComplete, pattern, title, onChange, onKeyDown, index } = props;
+	const { autoComplete, pattern, title, onChange, onKeyDown, index, value, maxLength, required } =
+		props;
 
 	const inputRef = useRef( null );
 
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
Works for me

outdoor2kode (approved):
Thanks for the fix, looks good 👍

outdoor2kode (approved):
Thanks for the fix, looks good 👍

Issue #158, opened 2023-05-11
Rest API shows backup codes enabled when TOTP disabled


iandunn:
Backup codes should be disabled when TOTP is disabled, per #75. The `2fa_available_providers` field will include Backup Codes even when it should be disabled, though.

1. Enable TOTP and backup codes
2. Disable TOTP

That problem is made a bit more obvious by #157 and https://github.com/WordPress/wporg-two-factor/pull/161#discussion_r1191885596. When TOTP/WebAuthn is disabled, our custom UI should show [the `Please enable Two-Factor Authentication before enabling backup codes` message](https://github.com/WordPress/wporg-two-factor/blob/d5a28af862cb8eb8aa7e7c7fb47d0343b619d8e1/settings/src/components/account-status.js#LL31C4-L31C4) in the Backup Codes card, but it currently doesn't. The `Your account has elevated privileges...` message should also be shown when TOTP/WebAuthn are disabled, but it isn't.

The potential problem was touched on briefly in https://github.com/WordPress/wporg-two-factor/pull/75#discussion_r1122118763, but I don't think we had a tangible problem until now. IIRC, switching to `two_factor_enabled_providers_for_user` should solve this issue while preserving the intent of #75

PR #161, opened 2023-05-12
Only show the reauth modal if a primary 2FA provider is enabled

### LINKED ISSUE #160: Fatal in reauth modal when 2FA is disabled
When 2FA has been enabled, and is then disabled, the user is still shown the modal to reauth, even though there is no 2FA provider available:

Screen Shot 2023-05-12 at 3 09 04 PM

Expected behaviour is for the modal not to be shown if 2FA is not enabled.

adamwoodnz:
Fixes #160 

This checks whether the `Two_Factor_Totp` is available before showing the reauth modal.

In the current state the backup codes are still accessible, but I believe the intention with https://github.com/WordPress/wporg-two-factor/issues/47 is to disable access to that screen as well if the `Two_Factor_Totp` provider is unavailable, see https://github.com/WordPress/wporg-two-factor/pull/157


settings/src/script.js
```diff
@@ -163,9 +163,8 @@ function Main( { userId } ) {
 		);
 	} else if (
 		twoFactorRequiredScreens.includes( screen ) &&
-		userRecord.record[ '2fa_available_providers' ] &&
-		userRecord.record[ '2fa_revalidation' ] &&
-		userRecord.record[ '2fa_revalidation' ].expires_at <= new Date().getTime() / 1000
+		userRecord.record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' ) &&
+		userRecord.record[ '2fa_revalidation' ]?.expires_at <= new Date().getTime() / 1000
```
adamwoodnz, line 167: If we're not going to disable the backup codes screen when totp is disabled, we'd need logic to check which provider is enabled, and show the relevant reauth UI; totp or backup code. Something like:

```
} else if (
		userRecord.record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' ) &&
		userRecord.record[ '2fa_revalidation' ]?.expires_at <= new Date().getTime() / 1000 
) {
                screenContent = (
			<>
				<div className={ 'wporg-2fa__account-status' }>
					<AccountStatus />
				</div>
				<RevalidateModalToTP />
			</>
		);
} else if (
		userRecord.record[ '2fa_available_providers' ].includes( 'Two_Factor_Backup_Codes' ) &&
		userRecord.record[ '2fa_revalidation' ]?.expires_at <= new Date().getTime() / 1000 
) {
                screenContent = (
			<>
				<div className={ 'wporg-2fa__account-status' }>
					<AccountStatus />
				</div>
				<RevalidateModalBackupCode />
			</>
		);
```

settings/src/script.js
```diff
@@ -156,22 +156,15 @@ function Main( { userId } ) {
...
-			</div>
-		);
+		screenContent = <AccountStatus />;
 	} else if (
 		twoFactorRequiredScreens.includes( screen ) &&
-		userRecord.record[ '2fa_available_providers' ] &&
-		userRecord.record[ '2fa_revalidation' ] &&
-		userRecord.record[ '2fa_revalidation' ].expires_at <= new Date().getTime() / 1000
+		record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' ) &&
+		record[ '2fa_revalidation' ]?.expires_at <= new Date().getTime() / 1000
```
adamwoodnz: If we're not going to disable the backup codes screen when totp is disabled, we'd need logic to check which provider is enabled, and show the relevant reauth UI; totp or backup code. Something like:

```javascript
const authExpired = record[ '2fa_revalidation' ]?.expires_at <= new Date().getTime() / 1000;

....

} else if (
		record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' ) && authExpired
) {
                screenContent = (
			<>
				<AccountStatus />
				<RevalidateModalToTP />
			</>
		);
} else if (
		record[ '2fa_available_providers' ].includes( 'Two_Factor_Backup_Codes' ) && authExpired
) {
                screenContent = (
			<>
				<AccountStatus />
				<RevalidateModalBackupCode />
			</>
		);
```

Might involve getting a second reauth URL from the server for backup codes too

pkevan:
While this would solve the current setup, I don't think we necessarily want to limit this to just TOTP since we are introducing webauthn too and that would require another fix here - is there any alternative method?

settings/src/script.js
```diff
@@ -156,22 +156,15 @@ function Main( { userId } ) {
...
-				<AccountStatus />
-			</div>
-		);
+		screenContent = <AccountStatus />;
 	} else if (
 		twoFactorRequiredScreens.includes( screen ) &&
-		userRecord.record[ '2fa_available_providers' ] &&
-		userRecord.record[ '2fa_revalidation' ] &&
-		userRecord.record[ '2fa_revalidation' ].expires_at <= new Date().getTime() / 1000
+		record[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' ) &&
```
pkevan: Can we achieve the same by checking the size of the array and that it doesn't just contain backup codes?

iandunn:
> I don't think we necessarily want to limit this to just TOTP since we are introducing webauthn too and that would require another fix here - is there any alternative method?

#157 introduces a `primaryProviderEnabled` bool in `AccountStatus`. If that'd be useful here too, we could maybe move it into `script.js`, and then add it to `GlobalContext` so `AccountStatus` can reuse it.

settings/src/script.js
iandunn: I think it will be disabled after #158 is resolved.

iandunn (commented):
I think the root cause for this is actually an upstream bug. I've opened https://github.com/WordPress/two-factor/issues/565 and https://github.com/WordPress/two-factor/pull/566 for that.

I think that will fix this, but haven't tested yet since it's still a WIP.

settings/src/script.js
dd32: > we'd need logic to check which provider is enabled, and show the relevant reauth UI; totp or backup code. 

This should really be handled by the Two-Factor handler I think rather than us (It should display whichever the user logged in with), but, we may need to extend the re-auth period when a login is using the Backup codes, as using up recovery codes every 10 minutes is not ideal.

Perhaps the time should be extended to 1hr for recovery code logins?

iandunn: I like that idea 👍🏻

settings/src/script.js
adamwoodnz: Updated

settings/src/script.js
```diff
@@ -130,6 +130,15 @@ function Main( { userId } ) {
...
 
+	const { '2fa_available_providers': availableProviders, '2fa_revalidation': revalidation } =
+		record;
+	// Check that there are providers, and that there isn't only backup codes.
+	const hasPrimaryProvider =
+		!! availableProviders.length &&
+		! (
+			availableProviders.length === 1 && availableProviders[ 0 ] === 'Two_Factor_Backup_Codes'
+		);
+
```
dd32: I'm wondering if it would be better to either a) have an array of "not valid" providers OR b) a list of primary providers.

Eg
```suggestion
	// Check that there are expected providers enabled.
	const primaryProviders = [ 'Two_Factor_Totp', 'TwoFactor_Provider_WebAuthn' ];
	const hasPrimaryProvider =
		!! availableProviders.filter( provider => primaryProviders.includes( provider ) ).length;

```

settings/src/script.js
```diff
@@ -156,22 +165,15 @@ function Main( { userId } ) {
...
 	if ( 'account-status' === screen ) {
-		screenContent = (
-			<div className={ 'wporg-2fa__account-status' }>
-				<AccountStatus />
-			</div>
-		);
+		screenContent = <AccountStatus />;
 	} else if (
 		twoFactorRequiredScreens.includes( screen ) &&
-		userRecord.record[ '2fa_available_providers' ] &&
```
dd32, line 166: 🤦  This is JS.. Even an empty array is truthful here.. this should've had a `.length` appended.. that's half the reason for this PR :D  (No changes needed, just adding a 🤯 moment)

settings/src/script.js
dd32: Otherwise, ignoring certain providers is probably also viable, but for testing on sandboxes, I'd recommend explicitly ignoring the Dummy method too, even though that shouldn't be seen in production.

```suggestion
	// Check that there are expected providers enabled.
	const ignoredProviders = [ 'Two_Factor_Backup_Codes', 'Two_Factor_Dummy' ];
	const hasPrimaryProvider =
		!! availableProviders.filter( provider => ! ignoredProviders.includes( provider ) ).length;

```

dd32 (approved):
This works for me, although I've added a few comments on the method of checking, since I think it'd be better to intersect arrays rather than checking for specific keys

settings/src/script.js
adamwoodnz: Yeah, just noticed @iandunn's [suggestion](https://github.com/WordPress/wporg-two-factor/pull/161#issuecomment-1546009919) about reusing the `primaryProviderEnabled` derived value from https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/account-status.js#L25

Having centralized logic for this probably makes sense.

dd32: Ahh yup, I didn't see that either. That makes sense to me :)

adamwoodnz: Updated to use a new boolean on `userRecord`

settings/src/utilities.js
```diff
@@ -18,6 +18,12 @@ export function useGetUserRecord( userId ) {
 		userRecord.record.password = '';
 	}
 
+	const availableProviders = userRecord.record?.[ '2fa_available_providers' ] ?? [];
+	const primaryProviders = [ 'Two_Factor_Totp', 'TwoFactor_Provider_WebAuthn' ];
+	userRecord.hasPrimaryProvider = !! availableProviders.filter( ( provider ) =>
+		primaryProviders.includes( provider )
+	).length;
+
```
iandunn: 🤔 This feels a bit unexpected to me, since it's mixing custom app data with a native Gutenberg object. `isSaving` is kind of doing that too, but it's more of a generic helper that Gutenberg will hopefully add one day.

A separate utility function like `hasPrimaryProvider( availableProviders ) {}` feels better to me. Or calculating it in `script.js`, and saving the result to `GlobalContext`. What do you think?

Not a big deal either way, and I don't feel strongly.

settings/src/components/tests/utlitites.test.js
```diff
@@ -55,6 +55,65 @@ describe( 'useGetUserRecord', () => {
...
+
+	it( 'should set hasPrimaryProvider to false if 2fa_available_providers is empty', () => {
+		useEntityRecord.mockReturnValue( { record: { '2fa_available_providers': [] } } );
+
+		const result = useGetUserRecord( 1 );
+
+		expect( result.hasPrimaryProvider ).toBe( false );
+	} );
+
+	it( 'should set hasPrimaryProvider to false if 2fa_available_providers only has Two_Factor_Backup_Codes', () => {
```
iandunn, line 83: Great job adding tests!

iandunn (approved):

settings/src/utilities.js
adamwoodnz: Yeah that's a fair call. I guess I saw the precedent and assumed we were treating it as our derived user state. I agree that mixing data isn't ideal. I'll rethink this 👍

StevenDufresne: I'm personally not too concerned about the derivation, but it since the language is vague, it could be open to future conflicts.

adamwoodnz: Restructured this so that we're leaving the `userRecord` unmodified, apart from the password override. Feels like we should be able to omit the mutation of the password value too, but I looked into it and it's not simple enough to be added to this PR.

iandunn: Doh, I missed that we were changing the password here too, that's a good point. IMO that has the same problem, but you're right that it establishes a precedent and changes the nature of this function.

[6a7290f](https://github.com/WordPress/wporg-two-factor/pull/161/commits/6a7290fd951baa118dbd84ccbf7b15359612204e) LGTM though 👍🏻

iandunn (approved):
LGTM 👍🏻

PR #164, opened 2023-05-16
Require an "ordinary" provider to use Backup Codes

### LINKED ISSUE #158: Rest API shows backup codes enabled when TOTP disabled

iandunn:
Fixes #158

This preserves the code from #75, but goes further to disable the provider for the user until they have an "ordinary" provider enabled.

"Ordinary" means that the provider is intended for normal use (like TOTP), as opposed to a fallback provider (like Backup Codes) that is only used when the user can't access their ordinary device.

## Testing

1. Enable TOTP and Backup Codes
2. Disable TOTP. You should see [the elevated privileges warning](https://github.com/WordPress/wporg-two-factor/blob/6a95fb3e72294a1d4a14c3a724958537dfc80259/wporg-two-factor.php#L250) on the front- and back-ends. The Backup Codes provider will also show as disabled on the front- and back-ends.

Screenshot 2023-05-22 at 2 45 41 PM

## Notes

* If WebAuthn is enabled and TOTP is disabled, the custom UI will still show "Two-Factor auth" (TOTP) as disabled. That will be fixed by #87 and subsequent PRs.
* This doesn't automatically update the `Two_Factor_Core::ENABLED_PROVIDERS_USER_META_KEY` database value, so, there are some normal situations where the two values are out of sync.

    For example, when a new user sets up TOTP and Backup Codes through the custom UI. In that scenario, the db value will only have TOTP, but in memory it'll have Backup Codes added. If there's a subsequent save to the value, then the database will be updated to reflect the value in memory.
    
    I don't think that'll cause any tangible issues in practice, but it's worth mentioning. If we want, we could add extra code to update the DB, but the time vs benefit tradeoff doesn't seem worth it to me at this point.

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -73,6 +73,11 @@ function render_custom_ui() : void {
 	echo do_blocks( "<!-- wp:wporg-two-factor/settings $json_attrs /-->" );
 }
 
+/**
+ * Print JS and CSS for customizing the interim revalidation iframe.
+ *
+ * @codeCoverageIgnore
+ */
 function login_footer_revalidate_customizations() {
 	// When the revalidate_2fa page is displayed in an interim login on not-login, add some style and JS handlers.
 	if (
--- a/tests/test-wporg-two-factor.php
+++ b/tests/test-wporg-two-factor.php
@@ -1,6 +1,6 @@
 <?php
 
-use function WordPressdotorg\Two_Factor\{ user_requires_2fa };
+use function WordPressdotorg\Two_Factor\{ user_requires_2fa, has_ordinary_provider };
 use function WordPressdotorg\MU_Plugins\Encryption\{ generate_encryption_key };
 
 defined( 'WPINC' ) || die();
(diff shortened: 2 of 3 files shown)


iandunn:
Nevermind @dd32 , @adamwoodnz . I'm switching this back to draft for now, to fix a test and also because it'll be hard to see the changes in the diff until #161 is merged.

iandunn:
@adamwoodnz , @dd32 ok, this is ready for review now.

PR #165, opened 2023-05-17
WebAuthN: Disable EdDSA/Ed25519 for key registrations

dd32:
Depends on: #153 
Patch for the upstream https://github.com/sjinks/wp-two-factor-provider-webauthn/issues/221 bug
See #114 

The upstream plugin supports a multitude of algorithms, unfortunately it appears that when a modern security key is registered through a desktop browser, EdDSA is preferred, however it appears that at least some Android devices do not support this through NFC.

This PR disables the EdDSA algorithm during key registration.

### DIFF
--- a/wporg-two-factor.php
+++ b/wporg-two-factor.php
@@ -185,7 +185,7 @@ function user_requires_2fa( $user ) : bool {
 			return false;
 		}
 	}
-    // @codeCoverageIgnoreEnd
+	// @codeCoverageIgnoreEnd
 
 	$required = false;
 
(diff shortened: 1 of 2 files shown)


wporg-two-factor.php
```diff
@@ -244,6 +244,42 @@ function get_edit_account_url() : string {
...
+ */
+add_action( 'wp_ajax_webauthn_preregister', __NAMESPACE__ . '\webauthn_preregister_remove_eddsa', 1 );
+function webauthn_preregister_remove_eddsa() {
+	ob_start( __NAMESPACE__ . '\webauthn_preregister_remove_eddsa_callback' );
+}
+
+/**
+ * Callback for webauthn_preregister_remove_eddsa().
+ *
+ * @codeCoverageIgnore
```
iandunn: It makes sense to me to ignore the `ob_start()` callback, but this one seems like a pure function. Are you ignoring it because it's temporary code and not worth testing long term?

iandunn (approved):
I had to merge the latest `trunk` in to make the WebAuthn provider show up in wp-admin, but after that it worked in my test 👍🏻

iandunn:
For reference, I tested the following scenarios and all of them worked:

* Yubikey 5C Nano: Registered via USB, authenticated via USB
* Yubikey 5C NFC: Registered via USB, authenticated via USB
* Yubikey 5C NFC: Registered via USB, authenticated via NFC

The desktop was a 2023 Mac Mini running OSX `13.4` and Chrome `113.0.5672.126`. The phone was a Pixel 4a running Android `13` and Chrome `113.0.5672.131`

wporg-two-factor.php
dd32: I'm not sure it matters either way, neither are likely to be unit tested IMHO, but both probably could be.

dd32:
I was thinking of merging #169 first and then applying this in the same manner, rather than having it in multiple places.

PR #166, opened 2023-05-18
Auto select TOTP input contents when focused

### LINKED ISSUE #155: TOTP Codes should auto-select any contents upon tab
1. Enter 123123 into the TOTP setup screen
2. Focus on the first box.
3. Type 4
4. Auto-tabbed to the next input
5. Type 5. Unable to, until you delete the character in the field.

Ideally at step 4, upon tabbing to an input that has contents, it should select the existing content to allow typing over it, but not clear it.

This could also be achieved as a general "Upon field being focused, select the character" style to allow easy overwrite.

adamwoodnz:
Closes #155 

This PR builds upon the paste handler added in #154

When an input is focused its contents are automatically selected, allowing the user to use Tab to move back and forth and type over previous values.

totp auto select

### How to test

Combinations of typing values, tabbing forwards and backwards, pasting, delete with backspace, clear all.

### DIFF
--- a/settings/src/components/auto-tabbing-input.js
+++ b/settings/src/components/auto-tabbing-input.js
@@ -9,27 +9,13 @@ import { useCallback } from '@wordpress/element';
 import NumericControl from './numeric-control';
 
 const AutoTabbingInput = ( props ) => {
-	const { inputs, setInputs, onComplete, error } = props;
+	const { inputs, setInputs, error } = props;
 
 	const handleChange = useCallback( ( value, event, index, inputRef ) => {
 		setInputs( ( prevInputs ) => {
 			const newInputs = [ ...prevInputs ];
 
-			// Clean input
-			if ( value.trim() === '' ) {
-				event.target.value = '';
-				value = '';
-			}
-
-			newInputs[ index ] = value;
-
-			// Check if all inputs are filled
-			const allFilled = newInputs.every( ( input ) => '' !== input );
-			if ( allFilled && onComplete ) {
-				onComplete( true );
-			} else {
-				onComplete( false );
-			}
+			newInputs[ index ] = value.trim() === '' ? '' : value;
 
 			return newInputs;
 		} );
@@ -51,7 +37,6 @@ const AutoTabbingInput = ( props ) => {
 		if ( inputs.length === paste.length ) {
 			event.preventDefault();
 			setInputs( paste.split( '' ) );
-			onComplete( true );
 		}
 	}, [] );
 
--- a/settings/src/components/auto-tabbing-input.js
+++ b/settings/src/components/auto-tabbing-input.js
@@ -46,17 +46,20 @@ const AutoTabbingInput = ( props ) => {
 	}, [] );
 
 	const handlePaste = useCallback( ( event ) => {
-		const paste = event.clipboardData.getData('Text').replace( /[^0-9]/g, '' )
+		const paste = event.clipboardData.getData( 'Text' ).replace( /[^0-9]/g, '' );
 
-		if ( inputs.length == paste.length ) {
+		if ( inputs.length === paste.length ) {
 			event.preventDefault();
-			setInputs( paste.split('') );
+			setInputs( paste.split( '' ) );
 			onComplete( true );
 		}
 	}, [] );
 
 	return (
-		<div className={ 'wporg-2fa__auto-tabbing-input' + ( error ? ' is-error' : '' ) } onPaste={ handlePaste }>
+		<div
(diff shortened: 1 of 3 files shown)


settings/src/components/auto-tabbing-input.js
```diff
@@ -11,26 +11,34 @@ import NumericControl from './numeric-control';
...
 		setInputs( ( prevInputs ) => {
 			const newInputs = [ ...prevInputs ];
 
 			newInputs[ index ] = value.trim() === '' ? '' : value;
 
 			return newInputs;
 		} );
+	}, [] );
 
-		if ( value && '' !== value.trim() && inputRef.current.nextElementSibling ) {
```
adamwoodnz, line 37: The change handler is now only responsible for updating the input values

settings/src/components/auto-tabbing-input.js
```diff
@@ -11,26 +11,34 @@ import NumericControl from './numeric-control';
...
-			inputRef.current.nextElementSibling.focus();
+	const handleKeyUp = useCallback( ( value, event, index, inputElement ) => {
```
adamwoodnz: Handling auto focus of the next/prev input `onKeyUp` means we have the new value, but aren't relying on whether the value has changed to determine whether to shift focus.

settings/src/components/numeric-control.js
```diff
@@ -13,27 +13,47 @@ import { useRef, useCallback } from '@wordpress/element';
...
-	const handleKeyDown = useCallback(
-		// Most callers will only need the value, so make it convenient for them.
-		( event ) => onKeyDown && onKeyDown( event.target.value, event, index, inputRef ),
-		[]
-	);
+	const handleChange = useCallback( createHandler( onChange ), [] );
+
+	const handleFocus = useCallback( createHandler( onFocus ), [] );
+
+	const handleKeyDown = useCallback( createHandler( onKeyDown ), [] );
```
adamwoodnz, line 54: This is no longer used. I've kept it as this seems designed to be a library component, but maybe it's YAGNI

settings/src/components/numeric-control.js
```diff
@@ -13,27 +13,47 @@ import { useRef, useCallback } from '@wordpress/element';
...
+} ) {
 	const inputRef = useRef( null );
 
-	const handleChange = useCallback(
+	const createHandler = ( handler ) => ( event ) =>
 		// Most callers will only need the value, so make it convenient for them.
-		( event ) => onChange && onChange( event.target.value, event, index, inputRef ),
-		[]
-	);
+		handler && handler( event.target.value, event, index, inputRef.current );
```
adamwoodnz, line 48: I don't think the consumer needs to know about the ref, we can just pass the `current` value back

dd32 (commented):
This works well in my testing, except, when multiple keys are pressed before the keyUp event happens.

For example, type `123123` by pressing 2 before 1 comes up, and 3 before 2 comes up and you'll see a "sluggish" UI that doesn't currently occur that misses some key presses.

Pastes still work as expected, and I'm not sure this is a real-life example that needs to be worried about, but you'll see the effect I mean if you try it..

| trunk | PR |
| --- | --- |
| <video src="https://github.com/WordPress/wporg-two-factor/assets/767313/a46b951b-f6f9-43ad-b7d4-aa4f0d99bd12"> | <video src="https://github.com/WordPress/wporg-two-factor/assets/767313/f8b82bd2-b5a4-4382-8b15-44bbedad3711"> |

adamwoodnz:
> I'm not sure this is a real-life example that needs to be worried about, but you'll see the effect I mean if you try it..

Yeah if you're a fast typer, and especially using a the numeric keypad, I can see it being potentially frustrating. Thanks, I'll see if I can improve it.

adamwoodnz:
> type `123123` by pressing 2 before 1 comes up, and 3 before 2 comes up and you'll see a "sluggish" UI that doesn't currently occur that misses some key presses

I think this should be fixed now @dd32

dd32:
@adamwoodnz Better :) I'm starting to question this though, because this is downright kind of crazy..

 - Can't overwrite first field
 - fast typing doesn't always overwrite
 - At this point this seems like this isn't reasonable to fix :) 

https://github.com/WordPress/wporg-two-factor/assets/767313/85cbaabe-cc16-44a9-b550-ff0fe78e21b1

adamwoodnz:
> * Can't overwrite first field
> * fast typing doesn't always overwrite

Hmm let me have another look

outdoor2kode:
> Can't overwrite first field
fast typing doesn't always overwrite

I've also run into these two issues. Additionally, there's another change that I find has become less convenient. Now when deleting digits, it needs continuously pressing backspace to tab backward, while originally it could be done by just holding the key down.

PR #169, opened 2023-05-19
WebAuthn changes required for production

dd32:
After #153 was merged, I attempted to enable it on WordPress.org, but ran into a number of problems.

 - The plugin doesn't handle shared user tables very well, and ended up creating tables such as `buddypress_2fa_webauthn_credentials` and would create `wordcamp_2fa_webauthn_credentials` too. Additionally it would run `dbDelta()` every load on those sites for some reason.
 - `Use WebAuthn` as the alternate provider text on login 'aint great; https://github.com/sjinks/wp-two-factor-provider-webauthn/pull/468
 - The WebAuthn plugin doesn't respect the revalidation process yet
 - The WebAuthn plugin doesn't employ caching, and calls to `::is_available_for_user()` trigger SQL queries on each call, this is mostly our problem as we're filtering the primary providers, so the additional caching is maybe irrelevant. Ref: https://github.com/WordPress/wporg-two-factor/pull/153#issuecomment-1550911721
 - After registering a new Security key, we should bump the session revalidation time. Ref: https://github.com/WordPress/wporg-two-factor/pull/153#issuecomment-1551038779

See #114


dd32:
> The WebAuthn plugin doesn't respect the revalidation process yet

Note: That's a two-pronged approach:
 - pre-ajax processing it aborts the request if the session isn't new enough
 - It wraps the UI with `<fieldset disabled>` which renders the buttons inoperable, but the links still load the disabled UI. Not pretty but works.

> The WebAuthn plugin doesn't employ caching

I'm not sure this is strictly required, but as I'd already written the code before I realised that, I included it here.

class-wporg-webauthn-provider.php
```diff
@@ -0,0 +1,131 @@
...
+
+	/**
+	 * See https://github.com/sjinks/wp-two-factor-provider-webauthn/pull/468
+	 *
+	 * @return string
+	 */
+	public function get_alternative_provider_label() {
+		return __( 'Use your security key', 'wporg-two-factor' );
+	}
+
```
iandunn, line 64:
```suggestion
```

https://github.com/sjinks/wp-two-factor-provider-webauthn/pull/468 has merged

iandunn (approved):
LGTM 👍🏻

class-wporg-webauthn-provider.php
dd32: Removing this makes sense to me, but depending on how long until the next upstream release happens it might be worth just keeping this to reduce UI feedback.

iandunn: That WFM 👍🏻

PR #174, opened 2023-05-23
Remove the Two_Factor profile actions from non-admin contexts only.

dd32:
After #153 we've got the Two Factor UI showing [up on the bbPress edit user screen](https://wordpress.org/support/users/profile/edit/), but the WebAuthn plugin doesn't support being displayed there so it doesn't show up.

This PR hides the Two Factor UI on the front-end again, and retains it for wp-admin contexts.

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -26,13 +26,15 @@ function register_block() {
  */
 function replace_core_ui_with_custom() : void {
 	/*
-	@todo Temporarily commented so that WebAuthn can be managed via wp-admin. Restore this when our custom WebAuthn UI is ready.
-	See https://github.com/WordPress/wporg-two-factor/issues/114, https://github.com/WordPress/wporg-two-factor/issues/87.
-	remove_action( 'show_user_profile', array( 'Two_Factor_Core', 'user_two_factor_options' ) );
-	remove_action( 'edit_user_profile', array( 'Two_Factor_Core', 'user_two_factor_options' ) );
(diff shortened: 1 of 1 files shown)


settings/settings.php
```diff
@@ -26,13 +26,15 @@ function register_block() {
...
 function replace_core_ui_with_custom() : void {
 	/*
-	@todo Temporarily commented so that WebAuthn can be managed via wp-admin. Restore this when our custom WebAuthn UI is ready.
-	See https://github.com/WordPress/wporg-two-factor/issues/114, https://github.com/WordPress/wporg-two-factor/issues/87.
-	remove_action( 'show_user_profile', array( 'Two_Factor_Core', 'user_two_factor_options' ) );
-	remove_action( 'edit_user_profile', array( 'Two_Factor_Core', 'user_two_factor_options' ) );
-	remove_action( 'personal_options_update', array( 'Two_Factor_Core', 'user_two_factor_options_update' ) );
-	remove_action( 'edit_user_profile_update', array( 'Two_Factor_Core', 'user_two_factor_options_update' ) );
-	*/
+	 * @todo Temporarily commented so that WebAuthn can be managed via wp-admin. Restore this when our custom WebAuthn UI is ready.
```
iandunn, line 29: I think the comment needs to be updated to reflect the new intent:

```diff
- * @todo Temporarily commented so that WebAuthn can be managed via wp-admin. Restore this when our custom WebAuthn UI is ready.
+ * @todo Temporarily scoped to the front end, so that WebAuthn can be managed via wp-admin. Remove the conditional when our custom WebAuthn UI is ready, so that the front-end is the only place to manage it.
```

Issue #172, opened 2023-05-22
Update local environments with WebAuthn plugin


iandunn:
See #114

After #153, some local environments like https://github.com/WordPress/wordcamp.org will fatal because the plugin is missing. We should make sure they're up to date with all the latest two-factor and wporg-two-factor related changes.

Props @ryelle for noticing

dd32:
I was thinking that we should probably also bail early from the loading function if the plugin isn't available.

Perhaps throwing a warning if the environment is `production`.

PR #176, opened 2023-05-23
Bail early if WebAuthn plugin missing.

dd32:
See #172


iandunn (approved):
I still think we should update the envs to install the plugin in order to close the issue, but this LGTM in the mean time 👍🏻

dd32:
I'm not against updating the environments too, I just think given this is an MU-plugin used everywhere, we should fail gracefully

PR #177, opened 2023-05-23
Fix dirty password prevent email from saving

### LINKED ISSUE #175: Dirty password field persists to email address view and prevents saving.
Related #117 

If you make an update to the password field, clear the field, then proceed to the email view, any changes there will return from the API as an invalid password, even though there are no password changes since the edited record has an empty password value.

We also need to clear the editedRecord of password changes if they are empty.

### Steps to reproduce
1. Visit password view, type in the field,
2. Navigate back
3. Navigate back into password view, notice the state is cleared
4. Navigate to the email address update view, notice the button is active (it shouldn't be)
5. Try changing your email address. You'll always get password errors from the api

## API Response

```
{"code":"rest_invalid_param","message":"Invalid parameter(s): password","data":{"status":400,"params":{"password":"Passwords cannot be empty."},"details":{"password":{"code":"rest_user_invalid_password","message":"Passwords cannot be empty.","data":{"status":400}}}}}
```
## Environment
MacOs Ventura, Chrome, wordpress.org

outdoor2kode:
Fixes #175

This PR fixes the issue that a dirty password prevents email from saving. A reassigning statement was added to the `script.js` to explicitly reset the password to its initial value.

## Screencast

https://github.com/WordPress/wporg-two-factor/assets/18050944/13e2d60d-f023-4959-97f1-f6d26d23b197

### DIFF
--- a/settings/src/script.js
+++ b/settings/src/script.js
@@ -115,14 +115,14 @@ function Main( { userId } ) {
 			event.preventDefault();
 
 			// Reset to initial after navigating away from a page.
+			// Note: password was initially not in record, this would prevent incomplete state
+			// from resetting when leaving the password setting page.
+			// See https://github.com/WordPress/wporg-two-factor/issues/117#issuecomment-1515693367.
 			if ( hasEdits ) {
-				edit( record );
-
-				// password is not reset by edit(record) as it's initially set as undefined.
-				// See https://github.com/WordPress/wporg-two-factor/issues/117#issuecomment-1515693367.
-				// we've tried setting its initial value as empty string, but this prevented email from being updated.
-				// See https://github.com/WordPress/wporg-two-factor/issues/175.
-				record.password = undefined;
+				edit( {
+					...record,
+					password: undefined,
+				} );
 			}
 
 			currentUrl = new URL( document.location.href );
--- a/settings/src/script.js
+++ b/settings/src/script.js
@@ -117,6 +117,12 @@ function Main( { userId } ) {
 			// Reset to initial after navigating away from a page.
 			if ( hasEdits ) {
 				edit( record );
+
+				// password is not reset by edit(record) as it's initially set as undefined.
+				// See https://github.com/WordPress/wporg-two-factor/issues/117#issuecomment-1515693367.
+				// we've tried setting its initial value as empty string, but this prevented email from being updated.
+				// See https://github.com/WordPress/wporg-two-factor/issues/175.
+				record.password = undefined;
 			}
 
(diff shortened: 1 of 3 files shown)


StevenDufresne:
Thanks. This appears to solve the issue for me 👍 . 

I'll leave the final review to @adamwoodnz as he's more familiar with the code.

settings/src/script.js
```diff
@@ -117,6 +117,12 @@ function Main( { userId } ) {
 			// Reset to initial after navigating away from a page.
 			if ( hasEdits ) {
 				edit( record );
+
+				// password is not reset by edit(record) as it's initially set as undefined.
+				// See https://github.com/WordPress/wporg-two-factor/issues/117#issuecomment-1515693367.
+				// we've tried setting its initial value as empty string, but this prevented email from being updated.
+				// See https://github.com/WordPress/wporg-two-factor/issues/175.
+				record.password = undefined;
```
adamwoodnz: Love the removal of the password mutation in `useUser`. I feel like using the `edit` method directly is cleaner, this works for me:

```suggestion
				// password is not reset by edit(record) as it's initially undefined.
				// See https://github.com/WordPress/wporg-two-factor/issues/117#issuecomment-1515693367.
				// we've tried setting its initial value as empty string, but this prevented email from being updated.
				// See https://github.com/WordPress/wporg-two-factor/issues/175.
				edit( {
				    ...record,
				    password: undefined,
				} );
```

settings/src/utilities.js
```diff
@@ -24,12 +24,6 @@ export function useUser( userId ) {
 		hasPrimaryProvider,
 	};
 
-	// Initialize the password as an empty string, necessary for resetting incomplete state when
-	// leaving the password setting page.
-	if ( user.userRecord.record && undefined === user.userRecord.record?.password ) {
-		user.userRecord.record.password = '';
-	}
-
 	return user;
```
adamwoodnz: We can return on line 21 now that the mutation is removed ❤️ 

```
        return {
		userRecord: { ...userRecord },
		isSaving,
		hasPrimaryProvider,
	};
```

settings/src/script.js
adamwoodnz: This also sets `hasEdits: false`, which wasn't happening with `record.password = undefined`

adamwoodnz (commented):
Nice work, very happy to clean up the `userUser` data! Couple of comments inline

settings/src/script.js
outdoor2kode: This is better, thanks!

PR #181, opened 2023-05-24
Remove fixed height so QR code can grow based on username length

### LINKED ISSUE #163: The barcode hover overlaps with copy.
In production, the hover state for the barcode control overlaps the copy.

URL: `https://wordpress.org/support/users/{username}/edit/account/?screen=totp`

## Screenshots
Screenshot 2023-05-15 at 3 32 00 PM

iandunn:
Fixes #163 

Otherwise a longer username would overflow the container and overlap surrounding elements.

See https://github.com/WordPress/two-factor/blob/654ebb75024c21c116b4155e7254e143c9f64413/providers/class-two-factor-totp.php#L234

### DIFF
--- a/settings/src/components/totp.scss
+++ b/settings/src/components/totp.scss
@@ -68,7 +68,6 @@
 
 	.wporg-2fa__qr-code {
 		text-align: center;
-		height: 225px;
 
 		#bbpress-forums & {
 			> a {


outdoor2kode (approved):
Thanks for the fix. It works for me. 👍

PR #178, opened 2023-05-23
Add gap between stacked password buttons on mobile

iandunn:
Fixes #162 

| Before | After |
|--------|--------|
| Screenshot 2023-05-15 at 3 26 11 PM | Screenshot 2023-05-23 at 3 35 44 PM | 

I added it to a few of the other providers who had similar layouts, but didn't include screenshots since they weren't affected.

PR #183, opened 2023-05-24
Combine action button container classes for DRYness

iandunn:
In #178 I missed that the TOTP component already had a class for action buttons. This combines this so that everything is consistent across components.

2e086d670cdec8cfcbff032a0ba8f453b231703f introduces a lot of diff noise, so I'd recommend reviewing w/ whitespace changes off. That change is necessary for 312a0a9fd567f3a59c6c110a72978499d663ccf8 to customize the general styles for TOTP.


outdoor2kode (approved):
Tested and all work as expected 👍

PR #184, opened 2023-05-25
Set default colors to legacy WP

### LINKED ISSUE #167: Incorrect color for interactive elements with Gutenberg 15.8
`--wp-admin-theme-color` seems to have changed to blueberry, and all the elements using it now don't match the rest of the UI

Screen Shot 2023-05-19 at 12 01 57 PM

Screen Shot 2023-05-19 at 12 05 43 PM

adamwoodnz:
Fixes #167 

This PR implements [the documented way of resetting the admin theme colors](https://github.com/WordPress/gutenberg/pull/50193) to be based on the legacy `#007cba` rather than Blueberry.

Currently it isn't working however, even though [`block.json` has `wp-components` as a dependency](--wp-admin-theme-color), the components styles are loaded last, overriding `--wp-admin-theme-color`.

### DIFF
--- a/settings/src/style.scss
+++ b/settings/src/style.scss
@@ -1,3 +1,5 @@
+@import '~@wordpress/base-styles/mixins';
+@import '~@wordpress/base-styles/default-custom-properties';
 @import '~@wordpress/base-styles/colors';
 @import '~@wordpress/base-styles/colors.native';
 


StevenDufresne:
@adamwoodnz We're moving this into profiles.wp.org, buddypress, which isn't a child theme of [wporg-parent-2021](https://github.com/WordPress/wporg-parent-2021) but has most of those styles. Can we close this Draft PR?

PR #185, opened 2023-05-25
Add workaround for reauth modal immediately closing in Firefox

### LINKED ISSUE #180: Can't revalidate with Firefox - iframe modal closes upon click
When the revalidation modal is triggered, it's not possible to click within the iframe part of it in Firefox.

This appears to be an upstream bug with the `useFocusableIframe` hook:
https://github.com/WordPress/gutenberg/issues/40912#issuecomment-1500233462

https://github.com/WordPress/wporg-two-factor/assets/767313/d0e1948e-ebab-4a35-b7c1-9aa6adc7f0c7

adamwoodnz:
Fixes #180 

This PR introduces the workaround mentioned on https://github.com/WordPress/gutenberg/issues/40912, which is to add `shouldCloseOnClickOutside={ false }` to the modal component, stopping the focus event in the iframe from closing the modal.

The downside is that clicking the modal background will not close the modal anymore, but the close button still works.

### Testing

**Ensure you use Firefox and other browsers.**

To force the modal to popup on the TOTP screen you can modify the [script.js condition](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/script.js#L166) to something like

```
else if ( twoFactorRequiredScreens.includes( screen ) ) {
    ...
```

### DIFF
--- a/settings/src/components/revalidate-modal.js
+++ b/settings/src/components/revalidate-modal.js
@@ -17,6 +17,9 @@ export default function RevalidateModal() {
 			title="Two-Factor Authentication"
 			onRequestClose={ goBack }
 			className="wporg-2fa__revalidate-modal"
+			// Temporary workaround until https://github.com/WordPress/gutenberg/issues/40912 is fixed.
+			// Without this the modal immediately closes in Firefox, see https://github.com/WordPress/wporg-two-factor/issues/180
+			shouldCloseOnClickOutside={ false }
 		>
 			<p>To update your two-factor options, you must first revalidate your session.</p>
 


dd32:
> The downside is that clicking the modal background will not close the modal anymore, but the close button still works.

IMHO I think this is a reasonable trade off, if/when the Gutenberg issue is resolved we can re-evaluate it.

dd32 (approved):
Tested with:
 - Safari 16.2
 - Firefox 114b8 (Beta)
 - Chrome 110 (Outdated)

Works for me, There's some differences between them like `ESC` works differently, sometimes it works until you focus into the iframe, etc.. but overall, it feels like things work as expected.

adamwoodnz:
Looks like [a fix for the Gutenberg issue was merged](https://github.com/WordPress/gutenberg/pull/51602) and will be released in v16.9

PR #187, opened 2023-05-25
Add try...catch to handle backup API failure

### LINKED ISSUE #182: UI should handle backup endpoint failure
Currently, if the REST API endpoint has a failure, the UI allow a user to confirm backup code have been generated and saved, confirming everything is fine.

https://github.com/WordPress/wporg-two-factor/assets/2290623/a3937003-1e47-4f06-ba50-6d2ac2d6f23b

Expected behaviour would be that upon a failed API response, or invalid data, that it is not possible to confirm anything has been saved.

outdoor2kode:
Fixes #182 

This PR adds try...catch to handle backup API failure, rendering an error notice when the API doesn't work correctly.

## Testing Steps

Go to `account-status.js` and comment out `disabled={ ! backupCodesEnabled }` on L78.

Scenario 1 - `error.code === 'revalidation_required'`

1. Go to `wp-content/plugins/two-factor/class-two-factor-core.php` and remove the `!` for the if condition on L1209.
=> `if ( self::current_user_can_update_two_factor_options( 'save' ) ) {`
2. Go to `/?screen=backup-codes` and it's expected to have the same results as the screenshot below.

Local

revalidation_require

Sandbox

sandbox

Scenario 2 - other kinds of error
1. Go to `wp-content/plugins/two-factor/providers/class-two-factor-backup-codes.php` and make it always throw an error on L295.
2. Go to `/?screen=backup-codes` and it's expected to have the same results as the screenshot below - an error notice with an error message is shown.

other errors

## Question
Do we want to show error messages dynamically on the front-end side, or alternatively, a certain error message like "Something went wrong, please seek assistance somewhere" could probably be displayed? As the error message like `"Unable to enable recovery codes for this user."` needs rephrasing a bit.

### DIFF
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -9,6 +9,7 @@ import { Icon, warning, cancelCircleFilled } from '@wordpress/icons';
 /**
  * Internal dependencies
  */
+import RevalidateModal from './revalidate-modal';
 import { GlobalContext } from '../script';
 import { refreshRecord } from '../utilities';
 
@@ -91,10 +92,13 @@ function Setup( { setRegenerating } ) {
 			<p>Please print the codes and keep them in a safe place.</p>
 
 			{ error ? (
-				<Notice status="error" isDismissible={ false }>
-					<Icon icon={ cancelCircleFilled } />
-					{ error.message }
-				</Notice>
+				<>
+					{ error.code === 'revalidation_required' && <RevalidateModal /> }
+					<Notice status="error" isDismissible={ false }>
+						<Icon icon={ cancelCircleFilled } />
+						{ error.message }
+					</Notice>
+				</>
 			) : (
 				<>
 					<CodeList codes={ backupCodes } />
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -4,7 +4,7 @@
 import apiFetch from '@wordpress/api-fetch';
 import { useContext, useCallback, useEffect, useState } from '@wordpress/element';
 import { Button, CheckboxControl, Notice, Spinner } from '@wordpress/components';
-import { Icon, warning } from '@wordpress/icons';
+import { Icon, warning, cancelCircleFilled } from '@wordpress/icons';
 
 /**
  * Internal dependencies
@@ -43,6 +43,7 @@ function Setup( { setRegenerating } ) {
 	} = useContext( GlobalContext );
 	const [ backupCodes, setBackupCodes ] = useState( [] );
 	const [ hasPrinted, setHasPrinted ] = useState( false );
+	const [ error, setError ] = useState( '' );
 
 	// Generate new backup codes and save them in usermeta.
 	useEffect( () => {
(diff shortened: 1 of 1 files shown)


iandunn:
> show error messages dynamically ... or a certain error message like "Something went wrong, please seek assistance somewhere"

IMO it's best to show a specific message, because that helps us troubleshoot when someone reports a bug. Otherwise it can be harder to reproduce. I don't feel strongly, though.

settings/src/components/backup-codes.js
```diff
@@ -85,20 +91,33 @@ function Setup( { setRegenerating } ) {
...
-			</Notice>
-
-			<CheckboxControl
-				label="I have printed or saved these codes"
-				checked={ hasPrinted }
-				onChange={ setHasPrinted }
-			/>
+			{ error ? (
+				<>
+					{ error.code === 'revalidation_required' && <RevalidateModal /> }
```
iandunn: 🤔 I feel like it might get hard to maintain if every component has to be aware of revalidation, and build in logic to handle it. I wonder if we can do something that'd covers all components, without them having to be coupled to it? That's probably outside the scope of this PR, though.

I'd lean towards removing the revalidation stuff here, and just having the `Notice` to catch any kind of API error. I don't feel strongly, though.

If you decide to keep the revalidation, then I think it'd be good to return the user to the Backup Codes screen after they revalidate, so they can pick up where they left off. Otherwise I worry the flow might feel disjointed to them.

iandunn (approved):
Looks good overall 👍🏻 

I think it's worth considering how revalidation is handled globally, but don't object to merging this either.

settings/src/components/backup-codes.js
StevenDufresne: Agreed. I would lean towards using an `apiFetch` middleware: https://www.npmjs.com/package/@wordpress/api-fetch.

pkevan:
> IMO it's best to show a specific message, because that helps us troubleshoot when someone reports a bug.

Agreed - always easier to have a specific place to start debugging from, rather than needing to reproduce :)

settings/src/components/backup-codes.js
outdoor2kode: Tried using `apiFetch` middleware, but it didn't go well. `Context` couldn't be accessed in the middleware and the RevalidateModal needs it. I've tried several other ways and found out that one seems to work with a quick test. I'll open another PR for it after finishing the rest part of it. And for this PR, I'll `remove the revalidation stuff here, and just have the Notice to catch any kind of API error`. Thanks for bringing this up, it's a really good point. And thanks Steve for sharing a potential solution! 👍

StevenDufresne: So I understand it correctly, you wanted to set the error within the app context and you couldn't?

outdoor2kode: [This](https://github.com/WordPress/wporg-two-factor/compare/trunk...try/revalidation-handler) is the branch I tried with `apiFetch` middleware, it's in the status that can be tested directly. I was trying to add a specific handler for revalidation using middleware, making it render the [components](https://github.com/WordPress/wporg-two-factor/compare/trunk...try/revalidation-handler#diff-ff88c76f4d9fc57f36f88dc481260c2d5ecd496149970a72162c81660f39fa2cR30-R37) we want if `'revalidation_required' !== error.code`. However, as `createRoot` or any other way that dynamically renders components would build a new tree that's unable to access the `Context` data, it would lead to an error message. This would work if only the component [Notice](https://github.com/WordPress/wporg-two-factor/compare/trunk...try/revalidation-handler#diff-ff88c76f4d9fc57f36f88dc481260c2d5ecd496149970a72162c81660f39fa2cR33-R36) is rendered. 

I also tried setting variables or flagging when the error occurs in the first place, and then having components detect them to trigger the re-rendering accordingly. However, this PR might be more [straightforward](https://github.com/WordPress/wporg-two-factor/pull/199).

PR #189, opened 2023-05-25
Fix unauthorized url access

### LINKED ISSUE #188: Users can still access backup-codes page through URL bar even if it's disabled
https://github.com/WordPress/wporg-two-factor/assets/18050944/fed5fcf5-d440-495d-ac73-774c6dd520cb

Expected behavior: 
1. Users shouldn't be able to access the backup-codes page if it's disabled. 
2. Status of the Two-Factor Backup Codes button shouldn't be updated to `check`.

outdoor2kode:
Fixes #188

This PR tidies up the code a bit for readability and DRYness and then makes sure users can't access the BackupCodes component via the URL bar when 2fa isn't enabled yet.

## Screencast

Sandbox

https://github.com/WordPress/wporg-two-factor/assets/18050944/5e4686e7-985b-48e8-a0bb-4e4c02dd3cba

### DIFF
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -84,9 +84,10 @@ function Setup( { setRegenerating } ) {
 	}, [] );
 
 	// Finish the setup process.
-	const handleFinished = useCallback( () => {
+	const handleFinished = useCallback( async () => {
+		// TODO: Add try catch here after https://github.com/WordPress/wporg-two-factor/pull/187/files is merged.
 		// The codes have already been saved to usermeta, see `generateCodes()` above.
-		refreshRecord( userRecord ); // This has the intended side-effect of redirecting to the Manage screen.
+		await refreshRecord( userRecord ); // This has the intended side-effect of redirecting to the Manage screen.
 		setGlobalNotice( 'Backup codes have been enabled.' );
 		setRegenerating( false );
 	} );
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -14,23 +14,18 @@ import { refreshRecord } from '../utilities';
 
 /**
  * Setup and manage backup codes.
- *
- * @param props
- * @param props.setScreen
  */
-export default function BackupCodes( { setScreen } ) {
+export default function BackupCodes() {
 	const {
-		user: { totpEnabled, backupCodesEnabled },
+		user: { backupCodesEnabled, totpEnabled },
+		navigateToScreen,
 	} = useContext( GlobalContext );
 	const [ regenerating, setRegenerating ] = useState( false );
 
 	// If TOTP hasn't been enabled, the user should not have access to BackupCodes component.
 	// This is primarily added to prevent users from accessing through the URL.
 	if ( ! totpEnabled ) {
-		const currentUrl = new URL( document.location.href );
-		currentUrl.searchParams.set( 'screen', 'account-status' );
-		window.history.pushState( {}, '', currentUrl );
-		setScreen( 'account-status' );
+		navigateToScreen( 'account-status' );
 		return;
 	}
 
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -14,12 +14,25 @@ import { refreshRecord } from '../utilities';
 
 /**
  * Setup and manage backup codes.
+ *
+ * @param props
+ * @param props.setScreen
  */
-export default function BackupCodes() {
+export default function BackupCodes( { setScreen } ) {
 	const {
-		user: { backupCodesEnabled },
+		user: { totpEnabled, backupCodesEnabled },
 	} = useContext( GlobalContext );
 	const [ regenerating, setRegenerating ] = useState( false );
+
+	// If TOTP hasn't been enabled, the user should not have access to BackupCodes component.
+	// This is primarily added to prevent users from accessing through the URL.
+	if ( ! totpEnabled ) {
+		const currentUrl = new URL( document.location.href );
+		currentUrl.searchParams.set( 'screen', 'account-status' );
+		window.history.pushState( {}, '', currentUrl );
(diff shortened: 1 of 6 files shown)


settings/src/script.js
```diff
@@ -60,31 +60,41 @@ function Main( { userId } ) {
 		hasPrimaryProvider,
 	} = user;
 	const [ globalNotice, setGlobalNotice ] = useState( '' );
+
+	const totpEnabled = record?.[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' );
+	const backupCodesEnabled =
+		record?.[ '2fa_available_providers' ].includes( 'Two_Factor_Backup_Codes' );
```
StevenDufresne: I'm generally in favor of moving these up the component chain or abstracting them into a shared state/context. I prefer that to how we currently reference them in the components, especially since they are provided by a third party plugin. Changing them in one place is better in my opinion and would provide a better api for graceful error handling. We are also not consistently checking for whether an object is defined: 
https://github.com/WordPress/wporg-two-factor/blob/cae02ef0aed8cc0ec9c9f162982a5d6078042461/settings/src/components/revalidate-modal.js#L49

With that being said, I do find it hard to follow which changes were need to fix the originating issue. Thoughts?

iandunn: I also find it harder to review PRs that do unrelated things, or get too big. What do you think about splitting those clean-up commits into a different PR?

I like the idea of pulling all the `{provider}Enabled`, `hasPrimaryProvider`,  etc things into the `Context` object 👍🏻

outdoor2kode: Thanks for pointing it out. I've split the DRYness commits into a separate [PR](https://github.com/WordPress/wporg-two-factor/pull/197), only related stuff is left in this one.

> I like the idea of pulling all the {provider}Enabled, hasPrimaryProvider, etc things into the Context object 👍🏻

IMHO, we might want to avoid using Context in this case as we're not dealing with `prop drilling` - passing data down through many levels of components, and also not dealing with passing down to _many_ components. Overusing Context sometimes might make the code hard to maintain.

outdoor2kode:
It has been rebased to trunk, and is ready for review again.

settings/src/components/backup-codes.js
```diff
@@ -14,13 +14,26 @@ import { refreshRecord } from '../utilities';
...
 	} = useContext( GlobalContext );
 	const [ regenerating, setRegenerating ] = useState( false );
 
+	// If TOTP hasn't been enabled, the user should not have access to BackupCodes component.
+	// This is primarily added to prevent users from accessing through the URL.
+	if ( ! totpEnabled ) {
+		const currentUrl = new URL( document.location.href );
+		currentUrl.searchParams.set( 'screen', 'account-status' );
+		window.history.pushState( {}, '', currentUrl );
+		setScreen( 'account-status' );
```
adamwoodnz: I actually had a go at doing just this recently and my approach was to reuse the internals of the `clickScreenLink` function, as this is almost exactly that. I created a branch which changed that to a basic navigation handler: https://github.com/WordPress/wporg-two-factor/compare/fix/unauthorized-backup-codes?expand=1

Events are handled at the component level and then `navigateToScreen` is called.

See what you think.

outdoor2kode: I like the enhancement 👍 Thanks for it.
I've cherry-picked the commits from your branch and tested the behavior on a sandbox, things worked as expected to me. Do you mind taking another look and reviewing again? Thanks.

iandunn:
#200 introduced dd26ab3a36d62b461ff18674db389ffe3a37f94e, so 26760b8ce1fa1b6e5d4f0049fca21d2f26b94525 no longer needs to modify `refreshRecord()`, it can just update the callers.

outdoor2kode:
> #200 introduced [dd26ab3](https://github.com/WordPress/wporg-two-factor/commit/dd26ab3a36d62b461ff18674db389ffe3a37f94e), so [26760b8](https://github.com/WordPress/wporg-two-factor/commit/26760b8ce1fa1b6e5d4f0049fca21d2f26b94525) no longer needs to modify `refreshRecord()`, it can just update the callers.

Thanks 👍  The code had been updated and so had the commit message. See f8abd296640198b3d9a84f426cda833afd407784.

settings/src/components/revalidate-modal.js
```diff
@@ -49,7 +52,13 @@ function RevalidateIframe() {
 			record[ '2fa_revalidation' ].expires_at = new Date().getTime() / 1000 + 3600;
 
 			// Refresh the user record, to fetch the correct 2fa_revalidation data.
-			refreshRecord( userRecord );
+			try {
+				await refreshRecord( userRecord );
+			} catch ( error ) {
+				// TODO: handle error more properly here, likely by showing a error notice
```
adamwoodnz, line 58: Maybe create an issue for this so we can follow up?

settings/src/script.js
```diff
@@ -63,14 +63,16 @@ function Main( { userId } ) {
...
+	const [ screen, setScreen ] = useState( initialScreen );
 
 	// The index is the URL slug and the value is the React component.
 	const components = {
 		'account-status': <AccountStatus />,
 		email: <EmailAddress />,
 		password: <Password />,
 		totp: <TOTP />,
-		'backup-codes': <BackupCodes />,
+		'backup-codes': <BackupCodes setScreen={ setScreen } />,
```
adamwoodnz: I don't think this is required anymore

adamwoodnz (commented):
This works for me 👍 

A couple of comments inline though

settings/src/components/revalidate-modal.js
outdoor2kode: Issue created https://github.com/WordPress/wporg-two-factor/issues/218.

settings/src/script.js
outdoor2kode: Thanks, missed this one.

Issue #194, opened 2023-05-26
Refine WebAuthn visual design


iandunn:
#95 stubbed in the rough code and UI, and #193 added real data/functionality.

In this PR we should refine that UI to match @jasmussen 's [design](https://github.com/WordPress/wporg-two-factor/pull/95#issuecomment-1563945641). I think Steve's idea of [a `Primary` indicator](https://github.com/WordPress/wporg-two-factor/pull/95#issuecomment-1563656017) is helpful too.

Canonical Figma:

* [checklist state 1](https://www.figma.com/file/6KPZbXujMGDRWWDR5odjDE/WP.org-improvements-%2F-QA?type=design&node-id=565-8992&t=Z1tueC4WkO24WnR5-4)
* [checklist state 2](https://www.figma.com/file/6KPZbXujMGDRWWDR5odjDE/WP.org-improvements-%2F-QA?type=design&node-id=565-12287&t=Z1tueC4WkO24WnR5-4)
* [security key flow](https://www.figma.com/file/6KPZbXujMGDRWWDR5odjDE/WP.org-improvements-%2F-QA?type=design&node-id=565-11747&t=Z1tueC4WkO24WnR5-4)
* [animation](https://www.figma.com/file/6KPZbXujMGDRWWDR5odjDE/WP.org-improvements-%2F-QA?type=design&node-id=565-12135&t=Z1tueC4WkO24WnR5-4)

iandunn:
Hey @jasmussen 👋🏻  I noticed a few differences between this and the TOTP flow:

* After setting up TOTP, the user is redirected to Backup Codes, but here they're redirected back to the Security Keys intro screen.
* TOTP adds a snackbar-esque notice on success, but this doesn't.
* TOTP has a button to disable it, but this doesn't.

Should the two screens be more consistent? What do you think about both screens redirecting to Backup Codes if they're not already setup, but back to the totp/security-key screen if backup codes are already setup?

jasmussen:
Any discrepancies are very likely to mainly be my lack of full understanding of the flow, so no reason to make too big of a detour. If you can provide me with screenshots of the flow, I'm happy to adjust, provide the remaining screens, or give feedback if perhaps we can optimize a part of it.

adamwoodnz:
> If you can provide me with screenshots of the flow

@jasmussen here are the differences

> TOTP has a button to disable it, but this doesn't.

[...]

PR #195, opened 2023-05-27
Delete WebAuthn key

### LINKED ISSUE #193: Enable WebAuthn UI functionality
#95 stubbed in the rough code and UI. In this PR we should add all the dynamic data and interactions. 

- [x] #196
- [x] #195
- [x] #204
- [x] #200 
- [x] #211 
- [x] Disable provider (without deleting keys, [pending discussion](https://github.com/WordPress/wporg-two-factor/issues/194#issuecomment-1564930700), depends on https://github.com/WordPress/two-factor/issues/585 

This doesn't need to do any UI refinement, though, that'll be done in #194.

iandunn:
See #193 

This adds the functionality for deleting a key. Design will be refined in #194, so this PR only focuses on the code. The Registration flow will be completed in #200, so this PR only focuses on the deletion flow.

Screenshot 2023-05-29 at 4 22 22 PM

<hr>

Screenshot 2023-05-29 at 4 22 42 PM

<hr>

Screenshot 2023-05-29 at 4 23 30 PM

<hr>

Screenshot 2023-05-29 at 4 23 57 PM

### DIFF
--- a/settings/src/components/webauthn/list-keys.js
+++ b/settings/src/components/webauthn/list-keys.js
@@ -1,38 +1,134 @@
 /**
  * WordPress dependencies
  */
-import { Button } from '@wordpress/components';
+import { Button, Modal, Notice, Spinner } from '@wordpress/components';
+import { useCallback, useContext, useState } from '@wordpress/element';
+import { Icon, cancelCircleFilled } from '@wordpress/icons';
 
-const confirm = window.confirm;
+/**
+ * Internal dependencies
+ */
+import { GlobalContext } from '../../script';
+import { refreshRecord } from '../../utilities';
+
+/**
+ * Global dependencies
+ */
+const ajax = wp.ajax;
 
 /**
  * Render the list of keys.
- *
- * @param props
- * @param props.keys
  */
-export default function ListKeys( { keys } ) {
-	return (
-		<ul>
-			{ keys.map( ( key ) => (
-				<li key={ key.id }>
-					{ key.name }
+export default function ListKeys() {
+	const {
+		user: { userRecord },
+		setGlobalNotice,
+	} = useContext( GlobalContext );
+	const keys = userRecord.record[ '2fa_webauthn_keys' ];
 
-					{ /* todo add onclick handler that pops up a <Modal> to confirm. maybe pass in from parent? */ }
-					<Button
-						variant="link"
-						data-id={ key.id }
-						aria-label="Delete"
-						onClick={ () =>
-							confirm(
-								'Modal H4 Remove Key? p Are you sure you want to remove the "" security key? Button Cancel Button Remove Key'
-							)
-						}
-					>
-						Delete
-					</Button>
-				</li>
-			) ) }
-		</ul>
+	const [ modalKey, setModalKey ] = useState( null );
+	const [ modalError, setModalError ] = useState( null );
+	const [ deleting, setDeleting ] = useState( false );
+
+	/**
+	 * After the user confirms their intent, POST an AJAX request to remove a key.
+	 */
+	const onConfirmDelete = useCallback( async () => {
+		setDeleting( true );
+
+		try {
+			await ajax.post( 'webauthn_delete_key', {
+				user_id: userRecord.record.id,
+				handle: modalKey.credential_id,
+				_ajax_nonce: modalKey.delete_nonce,
+			} );
+
+			setGlobalNotice( modalKey.name + ' has been deleted.' );
+			setModalKey( null );
+			refreshRecord( userRecord );
+		} catch ( error ) {
+			// The endpoint returns some errors as a string, but others as an object.
+			setModalError( error?.responseJSON?.data || error );
+		} finally {
+			setDeleting( false );
+		}
+	}, [ modalKey ] );
+
+	return (
+		<>
+			<ul>
+				{ keys.map( ( key ) => (
+					<li key={ key.id }>
+						{ key.name }
+
+						<Button
+							variant="link"
+							data-id={ key.id }
+							aria-label="Delete"
+							onClick={ () => setModalKey( key ) }
+						>
+							Delete
+						</Button>
+					</li>
+				) ) }
+			</ul>
+
+			{ modalKey && (
+				<ConfirmRemoveKey
+					keyToRemove={ modalKey }
+					error={ modalError }
+					deleting={ deleting }
+					onClose={ () => setModalKey( null ) }
+					onConfirm={ onConfirmDelete }
+				/>
+			) }
+		</>
+	);
+}
+
+/**
+ * Prompt the user to confirm they want to delete a key.
+ *
+ * @param {Object}   props
+ * @param {Object}   props.keyToRemove
+ * @param {Function} props.onConfirm
+ * @param {Function} props.onClose
+ * @param {boolean}  props.deleting
+ * @param {string}   props.error
+ */
+function ConfirmRemoveKey( { keyToRemove, onConfirm, onClose, deleting, error } ) {
+	return (
+		<Modal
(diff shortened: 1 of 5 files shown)


settings/src/components/webauthn/list-keys.js
```diff
@@ -1,38 +1,160 @@
...
+/**
+ * Internal dependencies
+ */
+import { GlobalContext } from '../../script';
+import { refreshRecord } from '../../utilities';
+
+/**
+ * Global dependencies
+ */
+const ajax = wp.ajax;
```
StevenDufresne, line 17: Why not use `apiFetch` the same as the other components:

https://github.com/WordPress/wporg-two-factor/blob/5fe83cf9453625836e4e82dfe8a608835ef901be/settings/src/components/totp.js#L50

settings/src/components/webauthn/list-keys.js
```diff
@@ -1,38 +1,160 @@
...
+ *
+ * @param {number}   userId
+ * @param {string}   credentialId
+ * @param {string}   nonce
+ * @param {Function} onSuccess
+ * @param {Function} onError
+ */
+async function handleRemoveKey( userId, credentialId, nonce, onSuccess, onError ) {
+	try {
+		await ajax.post( 'webauthn_delete_key', {
```
StevenDufresne: If you return the promise, you'll be able to use the default `then`, `catch`, `finally` and maybe not need this function.

`finally` is nice because you wouldn't need to make multiple calls to `setDeleting( false );`. You can call that once in `finally`.

settings/src/components/webauthn/webauthn.js
```diff
@@ -29,23 +31,27 @@ export default function WebAuthn() {
 	// and then replace userkeys with the value returned by the call
 	// probably also refreshRecord( userRecord );
 	const onRegisterSuccess = useCallback( () => {
-		userKeys.push( {
-			id: 1111,
+		const newKeys = userRecord.record[ '2fa_webauthn_keys' ].push( {
+			id: Math.random(),
```
StevenDufresne, line 35: Probably a very low risk of collisions?

settings/src/components/webauthn/list-keys.js
iandunn: This particular endpoint is admin-ajax, rather than the REST API. It's from the `two-factor-provider-webauthn` plugin, rather than one of our custom endpoints.

IIRC, using `apiFetch` would require a lot of customization since it's only intended to be used with REST endpoints. It might be easier to use native `fetch` if we wanted a modern JS API.

`wp.ajax` felt like like the best approach, though, since it's built specifically for admin-ajax endpoints.

settings/src/components/webauthn/webauthn.js
iandunn: Yeah, it's just a temporary placeholder to demo the functionality for Design. My next PR will replace it entirely with an API request that will create the credential in the database. At that point it'll have a unique ID enforced on the server side. 

I just used `Math.random()` here because a hardcoded value results in duplicate key warnings from React, since the component gets rendered multiple times.

settings/src/components/webauthn/list-keys.js
iandunn: That's a good point, i moved the func into `ListKeys` in b36783afae4338a7b250e9a9789881e5b0ecf887 so that it'd have direct access to the component members, rather than having to pass them in 👍🏻 

I also moved `setDeleting` to a `try...finally` block. 👍🏻 

At that point, the original `onConfirmDelete` function wasn't doing much, so I combined it with `handleRemoveKey`. I like this iteration a lot better, thanks! 🎉 

I still prefer to use `await`, since promise chaining has always felt very clunky to me. `await` feels conceptually simpler, and also much easier to read.

settings/src/components/webauthn/webauthn.js
outdoor2kode: Props/Context in React components should be considered read-only, it's best to avoid directly modifying props, or it could make the code harder to understand and maintain, and may lead to subtle bugs. What do you think if changing to something like the one below?

```js
const newKeys = [...userRecord.record[ '2fa_webauthn_keys' ], {id: Math.random(), name: 'New Key'}];
```

outdoor2kode:
Saw error messages when inputting. Has it been noticed and expected to be fixed in https://github.com/WordPress/wporg-two-factor/issues/194?

t

--

Also, I saw two keys always with the name `New Key` added when registering a new key while sandboxed or non-sandboxed, is it expected behavior at the moment?
image

--

And I couldn't Remove the key, it would get 400 error also on both sandbox and local env, is it expected at the moment as well or I might've missed some settings?
image

settings/src/components/webauthn/webauthn.js
iandunn: That's a good note about Context being read only 👍🏻  I tried that code, but it didn't result in the new item being added to the list.

I don't think we need to worry about it, though, because that section was just a stub from #95. It wasn't intended to  be used for more than a few days, and it'll be replaced entirely by #200.

iandunn:
> Saw error messages when inputting

I noticed that too, but the Registration flow is still just a stub from #95. The only reason I made a change to it in this PR was because I needed to decouple it from the Deletion flow (see https://github.com/WordPress/wporg-two-factor/pull/195/commits/5973b133fd63d7146800d664db1052b41b04948c)

The Deletion flow is what this PR is really about.

> two keys always with the name New Key added when registering a new key while sandboxed or non-sandboxed, is it expected behavior at the moment?

Yeah, that's just throwaway code from #95, it's only purpose is to demonstrate the flow to Design, and give us a code skeleton to iterate on. The deletion process should be working fully in this PR though.

> I couldn't Remove the key, it would get 400 error 

That's definitely a problem, but I haven't been able to reproduce it. Can you check that the `handle` and `
_ajax_nonce` fields in the `POST` request match their values in `wp-admin/profile.php`? Look for them in the `data-` attributes in `.webauthn-keys span.delete a`.

Also make sure that wp-admin isn't asking you to revalidate the 2fa session. Thanks!

settings/src/components/webauthn/webauthn.js
outdoor2kode: I see. Thanks for clarifying 👍

outdoor2kode:
> > Saw error messages when inputting
> 
> I noticed that too, but the Registration flow is still just a stub from #95. The only reason I made a change to it in this PR was because I needed to decouple it from the Deletion flow (see [5973b13](https://github.com/WordPress/wporg-two-factor/commit/5973b133fd63d7146800d664db1052b41b04948c))
> The Deletion flow is what this PR is really about.
> 
> > two keys always with the name New Key added when registering a new key while sandboxed or non-sandboxed, is it expected behavior at the moment?
> 
> Yeah, that's just throwaway code from #95, it's only purpose is to demonstrate the flow to Design, and give us a code skeleton to iterate on. The deletion process should be working fully in this PR though.

👍 

> > I couldn't Remove the key, it would get 400 error
> 
> That's definitely a problem, but I haven't been able to reproduce it. Can you check that the `handle` and ` _ajax_nonce` fields in the `POST` request match their values in `wp-admin/profile.php`? Look for them in the `data-` attributes in `.webauthn-keys span.delete a`.
> 
> Also make sure that wp-admin isn't asking you to revalidate the 2fa session. Thanks!

I see the problem. I thought I could directly register a new key using the button provided and then delete it, but then I realized that the Register Key functionality has not been implemented yet, as there's an open PR for it. I've added webauthn key manually and successfully deleted it on sandbox 👍 Sorry for the confusion.

iandunn:
Ah, that makes sense, my bad for not giving clearer testing instructions!

StevenDufresne (approved):

outdoor2kode (approved):

PR #197, opened 2023-05-29
Extract variables for DRYness

outdoor2kode:
This PR is separated from #189, aiming to enhance readability and avoid DRY.
It also extracts the webAuthnEnabled, which was added lately to trunk.

### DIFF
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -15,24 +15,20 @@ import ScreenLink from './screen-link';
  * Render the Account Status.
  *
  * @param root0
- * @param root0.backupCodesEnabled
  * @param root0.totpEnabled
+ * @param root0.backupCodesEnabled
+ * @param root0.webAuthnEnabled
  */
-export default function AccountStatus( { totpEnabled, backupCodesEnabled } ) {
+export default function AccountStatus( { totpEnabled, backupCodesEnabled, webAuthnEnabled } ) {
 	const {
 		user: {
 			userRecord: {
-				record: {
-					'2fa_available_providers': availableProviders,
-					email,
-					pending_email: pendingEmail,
-				},
+				record: { email, pending_email: pendingEmail },
 			},
 			hasPrimaryProvider,
 		},
 	} = useContext( GlobalContext );
 	const emailStatus = pendingEmail ? 'pending' : 'ok';
-	const webAuthnEnabled = availableProviders.includes( 'TwoFactor_Provider_WebAuthn' );
 
 	const backupBodyText =
 		! backupCodesEnabled && ! hasPrimaryProvider
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -13,8 +13,11 @@ import ScreenLink from './screen-link';
 
 /**
  * Render the Account Status.
+ *
+ * @param root0
+ * @param root0.backupCodesEnabled
  */
-export default function AccountStatus() {
+export default function AccountStatus( { backupCodesEnabled } ) {
 	const {
 		user: {
 			userRecord: {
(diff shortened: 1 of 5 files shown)


settings/src/components/account-status.js
```diff
@@ -13,24 +13,22 @@ import ScreenLink from './screen-link';
 
 /**
  * Render the Account Status.
+ *
+ * @param root0
+ * @param root0.totpEnabled
+ * @param root0.backupCodesEnabled
+ * @param root0.webAuthnEnabled
```
adamwoodnz: I know this is what we get with autofix, but I think `props` is more meaningful
```suggestion
 * @param props
 * @param props.totpEnabled
 * @param props.backupCodesEnabled
 * @param props.webAuthnEnabled
```

settings/src/script.js
```diff
@@ -61,20 +61,32 @@ function Main( { userId } ) {
 		hasPrimaryProvider,
 	} = user;
 	const [ globalNotice, setGlobalNotice ] = useState( '' );
+	const totpEnabled = record?.[ '2fa_available_providers' ].includes( 'Two_Factor_Totp' );
+	const backupCodesEnabled =
+		record?.[ '2fa_available_providers' ].includes( 'Two_Factor_Backup_Codes' );
+	const webAuthnEnabled = record?.[ '2fa_available_providers' ].includes(
+		'TwoFactor_Provider_WebAuthn'
+	);
```
adamwoodnz: I feel like passing these props down rather than getting them from the user context is a backward step. Maybe we should derive them there, similar to `hasPrimaryProvider`, and we can also add tests to make sure they handle the record being undefined.

utilities.js `useUser` would look something like:

```
const availableProviders = userRecord.record?.[ '2fa_available_providers' ] ?? [];

return {
	userRecord: { ...userRecord },
	isSaving,
	hasTotp: availableProviders.includes( 'Two_Factor_Totp' ),
	hasBackupCodes: availableProviders.includes( 'Two_Factor_Backup_Codes' ),
	hasWebAuthn: availableProviders.includes( 'TwoFactor_Provider_WebAuthn' ),
	hasPrimaryProvider: hasTotp || hasWebAuthn,
};
```

settings/src/components/account-status.js
outdoor2kode: Commit added: [752ec39](https://github.com/WordPress/wporg-two-factor/pull/197/commits/752ec39d821c186680af7fe4a28be9f6d6b5decd)

settings/src/script.js
outdoor2kode: That's a good idea. When referring to passing down these variables via `Context` in the first place, I was initially thinking of passing them down [here](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/script.js#L192), so I didn't think it would make things better. But placing it in `useUser` is a good choice! Thanks.

outdoor2kode: Commit pushed: [3c36cd9](https://github.com/WordPress/wporg-two-factor/pull/197/commits/3c36cd9b004ff706cd02a238bacb4f3342dc4399)

settings/src/utilities.js
```diff
@@ -17,11 +17,17 @@ export function useUser( userId ) {
 	const hasPrimaryProvider = !! availableProviders.filter( ( provider ) =>
```
adamwoodnz: This loop with inner loop is potentially expensive, and we already have the answer from `totpEnabled` and `webAuthnEnabled`. Suggest replacing with `const hasPrimaryProvider = totpEnabled || webAuthnEnabled;`.

In theory the tests should still pass.

outdoor2kode: Commit added: [90306c5](https://github.com/WordPress/wporg-two-factor/pull/197/commits/90306c546408bb481b2ea40e1902d5d11fb2a5d3)

adamwoodnz:
Hmm doesn't look like we're running JS tests on PRs

outdoor2kode:
> Hmm doesn't look like we're running JS tests on PRs

Hmm, I hadn't realized it until you pointed it out. Turns out only the PHP tests are being run.
I've tested it locally and tests are passed like you said, though.

outdoor2kode:
> Hmm doesn't look like we're running JS tests on PRs

Issue added: https://github.com/WordPress/wporg-two-factor/issues/201

adamwoodnz (approved):
LGTM :shipit:

PR #199, opened 2023-05-30
Handle revalidation error in a single place

outdoor2kode:
Fixes https://github.com/WordPress/wporg-two-factor/pull/187#discussion_r1206073166

## Testing Step

1. Go to `account-status.js` and comment out `disabled={ ! backupCodesEnabled }` on L78.
2. Go to `wp-content/plugins/two-factor/class-two-factor-core.php` and remove the `!` for the if condition on L1209.
=> `if ( self::current_user_can_update_two_factor_options( 'save' ) ) {`
3. Comment out the following snippet in `/wp-content/plugins/wporg-two-factor/settings/src/components/backup-codes.js` 
```
if ( ! totpEnabled ) {
	const currentUrl = new URL( document.location.href );
	currentUrl.searchParams.set( 'screen', 'account-status' );
	window.history.pushState( {}, '', currentUrl );
	setScreen( 'account-status' );
	return;
}
```
4. You will see the Modal popup when visiting `/?screen=backup-codes`. If you want to observe the same behavior as in the clip below, please use sandbox and make sure you have had 2fa set up, and then do `2.` above.

https://github.com/WordPress/wporg-two-factor/assets/18050944/9d12e948-c7bd-4f8e-a300-d8f6019cf610

## TODO

If this method works, pulling up all the `setError` and `error` in child components to `script.js`

### DIFF
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -46,10 +46,11 @@ function Setup( { setRegenerating } ) {
 	const {
 		setGlobalNotice,
 		user: { userRecord },
+		setError,
+		error,
 	} = useContext( GlobalContext );
 	const [ backupCodes, setBackupCodes ] = useState( [] );
 	const [ hasPrinted, setHasPrinted ] = useState( false );
-	const [ error, setError ] = useState( '' );
 
 	// Generate new backup codes and save them in usermeta.
 	useEffect( () => {
--- a/settings/src/script.js
+++ b/settings/src/script.js
@@ -72,7 +72,7 @@ function Main( { userId } ) {
 		email: <EmailAddress />,
 		password: <Password />,
 		totp: <TOTP />,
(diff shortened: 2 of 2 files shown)


adamwoodnz (commented):
This works for me, worth trying to use `setError` everywhere I think. We might need to take care of clearing the error centrally when navigating [like we do with the `globalNotice`](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/script.js#L138)

iandunn (approved):
Looks good :)

outdoor2kode:
> We might need to take care of clearing the error centrally when navigating [like we do with the globalNotice](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/script.js#L138)

Good catch! a `setError` should be added there. Added in 494358fd06d1b21329e9fde1092fa944594b7ef0.

Note: Should be merged after #189.

PR #200, opened 2023-05-30
Register WebAuthn keys

iandunn:
See #193

This adds the functionality to register WebAuthn keys. It focuses on the functionality only, since refining the design will be covered in #194. Enabling/disabling the provider will also be done in subsequent PRs.

Screenshot 2023-06-05 at 1 40 40 PM
<hr>
Screenshot 2023-06-05 at 1 40 54 PM
<hr>
Screenshot 2023-06-05 at 1 41 15 PM
<hr>
Screenshot 2023-06-05 at 1 41 41 PM
<hr>
Screenshot 2023-06-05 at 1 43 10 PM


iandunn:
I think we'll be able to re-use the `two-factor-provider-webauthn` AJAX endpoints here, like we did in #195. Unlike that PR, though, there's a non-trivial amount of work involved in the client-side code:

https://github.com/sjinks/wp-two-factor-provider-webauthn/blob/a17ed697a218741986198366217d4e73daa4403f/assets/profile.ts#L93-L139

https://github.com/sjinks/wp-two-factor-provider-webauthn/blob/a17ed697a218741986198366217d4e73daa4403f/assets/common.ts#L41-L111

I thought of a few different ways to approach this:

1. Write our own client side code to call their PHP APIs, like #195. This seems like it'd be more work (and take longer) than is necessary, though.
2. Call their JS code from our JS code
	This won't work, because theirs is transpiled into an IIFE

	https://github.com/sjinks/wp-two-factor-provider-webauthn/blob/a17ed697a218741986198366217d4e73daa4403f/rollup.config.mjs#L23-L27
3. Some kind of `node` process where we take in their code and wrap it in something that makes it accessible to us, so we can do `2`. Maybe adding the plugin to npm `devDependencies`, and then `import`ing it in `webauthn.js`. I think that'll need some custom Webpack config though. We could maybe run their build process, but override the rollup config w/ cli options, like changing `format: 'iife'` to `es`. See https://rollupjs.org/configuration-options/#output-format.
4. Submit a PR upstream to make any changes needed to facilitate `2` or `3`. Nothing jumped out at me as something that'd make sense for both them and us, though.
5. Fork and customize their code. This seems simpler than dealing w/ Webpack/rollup, and maybe less fragile when the upstream plugin changes over time. As long as the PHP API remains compatible, we wouldn't have to worry about JS changes.

I'm thinking `5` is the best approach, so I'm going to start testing that out.

iandunn:
It looks like @renintw and I both independently arrived at the need for returning the promise from `refreshRecord()`. His version is in https://github.com/WordPress/wporg-two-factor/pull/189/commits/411dcbf45bd174b5b44fa7848ef54cdfef020751, and mine is in [3406367](https://github.com/WordPress/wporg-two-factor/pull/200/commits/3406367297dcb5e6b989d0b39a41405b8150300f).

I avoided updating the other components in this PR in order to keep it focused on WebAuthn, but like the idea of updating the other components in a subsequent PR.

settings/src/components/webauthn/sjinks-utilities.js
```diff
@@ -0,0 +1,78 @@
+/**
```
outdoor2kode: Do you think we can move this section of the code into utilities? This way, we can keep only components under the component directory for consistency. 

Or maybe we could add a directory called `utilities` and put this file there, and then could also rename the original `utilities.js` to `index.js`/`common.js` to indicate its generic use? What do you think?

--

And I actually think the customized hook `useUser` currently in utilities is the best to be moved to a new directory called hooks, and be placed in a file called `useUser`. We can change this in another PR, though.

settings/src/components/webauthn/register-key.js
```diff
@@ -1,84 +1,109 @@
...
+import { preparePublicKeyCreationOptions, preparePublicKeyCredential } from './sjinks-utilities';
+import { refreshRecord } from '../../utilities';
 
 /**
  * Render the form to register new security keys.
  *
  * @param {Object}   props
- * @param {Function} props.onRegisterSuccess
+ * @param {Function} props.onCancel
+ * @param            props.onSuccess
```
outdoor2kode: I remember only TypeScript can automatically infer from the context that this field is a {Function}, I might be wrong, though.

settings/src/components/webauthn/register-key.js
```diff
@@ -1,84 +1,109 @@
...
-		<form>
-			<TextControl label="Give the security key a name" />
-			{ /* TODO add basic clientside validation
-			what? lendth? chars?
-			TextControl doesn't have anything special, just pass `maxlength`, `pattern`, etc
-			*/ }
+		<form onSubmit={ onRegister }>
+			<TextControl
+				label="Give the security key a name"
+				onChange={ ( name ) => setKeyName( name ) }
```
outdoor2kode, line 84: Not sure if an empty keyName is allowed, for now it would be `NewKey` if nothing is input.
Probably some checks on front-end side could be added in https://github.com/WordPress/wporg-two-factor/issues/194.

outdoor2kode (approved):
Functionality tested and looks good to me 👍 Only a few minor suggestions.

outdoor2kode (approved):
Functionality tested and looks good to me 👍 Only a few minor suggestions.

settings/src/components/webauthn/register-key.js
iandunn: Yeah, the API endpoint sets that default string here:

https://github.com/sjinks/wp-two-factor-provider-webauthn/blob/32c8c10f3f67a40a87e3a5f0cd643db568343bf7/inc/class-ajax.php#L121-L138

But I think it makes sense to to add a `required` attribute on our side. Done in ff14ce1 👍🏻

settings/src/components/webauthn/register-key.js
iandunn: I'm not sure, IIRC this is just what Prettier added automatically. Do you think there's a better value?

iandunn: There is `@callback`, but that seems kind of excessive, since you have to add a docblock at the function definition in addition to the `@param` here. 

That's not a big deal for things like `onRegisterSuccess` where we already have a docblock, but if we wanted to be consistent then we'd also need to do it for small closures like `onCancel`:

https://github.com/WordPress/wporg-two-factor/blob/d633ba92c06f372bc84f8d8f2215010ec6675eae/settings/src/components/webauthn/webauthn.js#L77

settings/src/components/webauthn/sjinks-utilities.js
iandunn: I think it's best to keep it in a separate file from our 1st-party code, to have a clean separation between that and our forked/3rd party code. I also like a separate file because of the length and logical grouping of the functions.

It feels to me like it fits better in the `webauthn` folder, since it's only relevant to that provider. I don't feel strongly though, and a new `utilities` folder works for me too 👍🏻 

In that case, should we also move `tests` outside of `components` (in another PR)?

outdoor2kode: > In that case, should we also move tests outside of components (in another PR)?

Ah, yep, I believe it's better to do so as I can see that there's a test for `utilities` which is in the upper level.

settings/src/components/webauthn/register-key.js
outdoor2kode: I was only trying to point out that the param type is empty for `props.onSuccess`, like `onCancel`, it's supposed to be `{Function}` as well. This is really minor stuff, in case one day we want to leverage the JSDoc with some tools, and normally the empty here would make tools not able to parse correctly.

But thanks for going in-depth for this 👍

iandunn: Ah, I missed that, thanks! Added in ff3684a 👍🏻

settings/src/components/webauthn/sjinks-utilities.js
iandunn: 👍🏻 I moved the file in 5468a00 and 0f201fd, and opened #210 for doing the out-of-scope changes

outdoor2kode: Thanks!

PR #205, opened 2023-05-31
Fix backup codes screen status not updated if go back without checkbox confirmed

### LINKED ISSUE #190: Handle the case that If "I have printed or saved these codes" in backupCodes is not clicked
At the moment, if a user doesn't click "I have printed or saved these codes" after backup codes are generated,
and directly go back to account-status screen, the enable status isn't correctly updated and reflected.

I have printed not clicked

outdoor2kode:
Fixes #190

This PR ensures that users cannot leave the Backup Codes screen by pressing `Back` without checking `I have printed or saved these codes`, and at the same time, prompt a Notice to remind users that they should confirm the checkbox before proceeding with other actions.

[ScreenNavigation was extracted](https://github.com/WordPress/wporg-two-factor/pull/205/commits/22b3e6c234142eea2724f4adb5b2fff2fedfeeb1) was done to make the subsequent commits easier to modify and read.

[hasPrinted was extracted to useUser](https://github.com/WordPress/wporg-two-factor/pull/205/commits/27ed028ecc030eac93e74fbf012098c1a0912a42) so that `navigateToScreen` can access it.

## Screencast

No other errors in BackupCodes component exist

https://github.com/WordPress/wporg-two-factor/assets/18050944/d9de3af7-e325-48a7-a37f-ce9f6fd03091

Other errors in BackupCodes component exist

https://github.com/WordPress/wporg-two-factor/assets/18050944/75dc36b1-508e-4121-bc78-b5bb5f7c05fc

## Testing Steps

1. Test in a Sandbox.
2. enable 2fa.
3. Go to backup codes and you can't `Back` if you don't confirm the checkbox, and an error Notice should prompt.
4. Go to `wp-content/plugins/two-factor/providers/class-two-factor-backup-codes.php` and make it always throw an error on L295.
5. Go to backup codes and you can click `Back`.

### DIFF
--- a/settings/src/script.js
+++ b/settings/src/script.js
@@ -114,10 +114,8 @@ function Main( { userId } ) {
 	 * This is used in conjunction with real links in order to preserve deep linking and other foundational
 	 * behaviors that are broken otherwise.
 	 */
-	const clickScreenLink = useCallback(
-		( event, nextScreen ) => {
-			event.preventDefault();
-
+	const navigateToScreen = useCallback(
+		( nextScreen ) => {
 			// Reset to initial after navigating away from a page.
 			// Note: password was initially not in record, this would prevent incomplete state
 			// from resetting when leaving the password setting page.
@@ -185,7 +183,7 @@ function Main( { userId } ) {
 	}
 
 	return (
-		<GlobalContext.Provider value={ { clickScreenLink, user, setGlobalNotice } }>
+		<GlobalContext.Provider value={ { navigateToScreen, user, setGlobalNotice } }>
 			<GlobalNotice notice={ globalNotice } setNotice={ setGlobalNotice } />
 			{ screenContent }
 		</GlobalContext.Provider>
--- a/settings/src/script.js
+++ b/settings/src/script.js
@@ -58,6 +58,8 @@ function Main( { userId } ) {
 	const {
 		userRecord: { record, edit, hasEdits, hasResolved },
 		hasPrimaryProvider,
+		hasBackupCodesPrinted,
+		setHasBackupCodesPrinted,
 	} = user;
 	const [ globalNotice, setGlobalNotice ] = useState( '' );
 	const [ error, setError ] = useState( '' );
@@ -165,7 +167,15 @@ function Main( { userId } ) {
 
 	return (
 		<GlobalContext.Provider
-			value={ { navigateToScreen, user, setGlobalNotice, setError, error } }
+			value={ {
+				navigateToScreen,
+				user,
+				setGlobalNotice,
+				setError,
+				error,
+				hasBackupCodesPrinted,
+				setHasBackupCodesPrinted,
+			} }
 		>
 			<GlobalNotice notice={ globalNotice } setNotice={ setGlobalNotice } />
 			{ currentScreenComponent }
--- a/settings/src/script.js
+++ b/settings/src/script.js
@@ -121,15 +121,21 @@ function Main( { userId } ) {
 	 */
 	const navigateToScreen = useCallback(
 		( { currentScreen, nextScreen } ) => {
-			// If there's already another error, do not overwrite it with the error below.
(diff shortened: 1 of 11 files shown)


settings/src/script.js
```diff
@@ -182,10 +180,18 @@ function Main( { userId } ) {
...
 		<GlobalContext.Provider
-			value={ { navigateToScreen, user, setGlobalNotice, setError, error } }
+			value={ {
+				navigateToScreen,
+				user,
+				setGlobalNotice,
+				setError,
+				error,
+				hasBackupCodesPrinted,
+				setHasBackupCodesPrinted,
```
iandunn: Do these need to be added here?

It seems like they're already available under `user`, similar to `hasPrimaryProvider`, `backupCodesEnabled`, etc.

iandunn (commented):
🤔 I have some reservations about this approach. 

It seems like there's two things going on here:
1. The superficial issue is that the UI isn't updated automatically to reflect the current state after calling `generatedCodes()`
2. The deeper issues is that the upstream plugin [implicitly saves/enables backup codes to the database before the user intends](https://github.com/WordPress/two-factor/issues/507), but that's not how the design was created.
 
IMO, the design is correct, and the codes shouldn't be active until the user confirms they want to save them. That gives the user a choice whether or not they want to activate them, and makes sure they're aware that the codes are active. 

This PR does make the user aware that they're active, but I worry that forcing them to stay on the screen  will feel like a frustrating UX. It also doesn't achieve the goal of giving them a choice to cancel. Code-wide, I feels like this may add more complexity than we need. 

I thought of a few potential alternatives:

1. Fix the UI state mismatch by calling `refreshRecord()` [after the codes are generated](https://github.com/WordPress/wporg-two-factor/blob/2321b3420d836b38c761b1b62b816f1b45f10fb1/settings/src/components/backup-codes.js#L46-L71)

    ```diff
    -generateCodes();
    +await generateCodes();
    +await refreshRecord();
    ```

1. Remove the `confirm` checkbox from the UI for now, and replace it with a message indicating the codes are active. Then when https://github.com/WordPress/two-factor/issues/507 is fixed, we could bring it back.
1. Write our own API endpoint to generate backup codes, instead of calling the upstream API from the client. It could [call the upstream endpoint internally](https://developer.wordpress.org/rest-api/requests/#internal-requests), and then disable the backup codes immediately after. Then, when the user clicks `confirm`, our client would send another XHR to enable them.

If `#1` would work, then that feels to me like the simplest option. We could also consider `#2` if we want to go further. 

I don't feel strongly though. What do you think @renintw and @adamwoodnz ?

settings/src/script.js
outdoor2kode: Ah, missed this one. They should be removed. d03c91ec5e940528eadffb4434db24e0959aa1f7

outdoor2kode:
> but I worry that forcing them to stay on the screen will feel like a frustrating UX

I guess I wasn't aware of the case that if a user had already activated and saved a set of backup-codes, and then accidentally clicked the generate new backup codes button, or clicked it intentionally but then decided they didn't want new codes, this would be a frustrating UX like you said. In this case, the issue shouldn't be addressed by approach in this PR.

I'll open new PRs for #1 and #2, and extract the refactoring part in this PR that I think might be worth retaining.

Thanks for the feedback 🙏 

**Update**

The problem seems to be solved in a different way in #217 and once https://github.com/WordPress/two-factor/issues/507 is fixed, this issue `The user should be able to enter the screen to see the status and regenerate codes, though.` mentioned [here](https://github.com/WordPress/wporg-two-factor/issues/190#issuecomment-1564616358) could be fixed by changing this [Line](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/account-status.js#L88) to `disabled={ ! backupCodesEnabled && ! totpEnabled}`

For `extract the refactoring part`, please see https://github.com/WordPress/wporg-two-factor/pull/215.

outdoor2kode:
Closed. See #217 for potential solution.

PR #207, opened 2023-06-01
Fix JS test failure due to undefined `wp`

### LINKED ISSUE #206: JS tests failing on trunk due to undefined `wp`
This line in [list-keys.js](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/webauthn/list-keys.js#L17) is causing the failure

````
/**
 * Global dependencies
 */
const ajax = wp.ajax;
```

`wp` is an eslint allowed global but the tests don;t seem to recognise this.

adamwoodnz:
Fixes #206

Remove reference to `wp` outside of component to fix undefined in password.test.js

### DIFF
--- a/settings/src/components/webauthn/list-keys.js
+++ b/settings/src/components/webauthn/list-keys.js
@@ -12,11 +12,6 @@ import { GlobalContext } from '../../script';
 import { refreshRecord } from '../../utilities';
 
 /**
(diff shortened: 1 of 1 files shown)


iandunn (approved):
This WFM 👍🏻

PR #211, opened 2023-06-07
Enable WebAuthn when registering the first key

iandunn:
See #193 

This enables the WebAuthn provider after a user registers their first key. It focuses on the functionality since the design will be refined in #194.

### Testing:

1. Disable WebAuthn provider on `wp-admin/profile.php`
2. Register a key
3. Refresh `profile.php` to see that it's now enabled

### DIFF
--- a/settings/rest-api.php
+++ b/settings/rest-api.php
@@ -3,7 +3,7 @@
 namespace WordPressdotorg\Two_Factor;
 use Two_Factor_Core, Two_Factor_Totp, Two_Factor_Backup_Codes;
 use WildWolf\WordPress\TwoFactorWebAuthn\{ WebAuthn_Credential_Store };
-use WP_REST_Server, WP_REST_Request;
+use WP_REST_Server, WP_REST_Request, WP_Error, WP_User;
 
 defined( 'WPINC' ) || die();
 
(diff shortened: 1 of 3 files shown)


StevenDufresne (approved):
This worked as expected. 👍 

I don't have any issues with this merging but maybe a second opinion about the rest usage may be useful.

dd32:
I haven't tested, and have no real issue with the approach, but part of me wonders if this is the right way to go about it.

Since we're filtering the enabled providers for users already, we could probably just have it enabled for every user?  The status is then based on whether or not the user has any keys registered.

Relying upon a second API request after registering the key just feels like it could be potentially fragile?

iandunn:
Yeah, that would make the UI and code simpler. I didn't go that way because it'd prevent folks from disabling the provider while keeping the keys registered. 

One situation where that could be useful is if you want to register passkeys (or some other non-portable security key) on multiple devices before you enable 2FA, so that you don't have to use a backup code when logging in on the 2nd device.

That may be an edge case, though. I don't feel strongly.

iandunn:
I'm gonna go ahead and merge this so it doesn't fall through the cracks, but I'm definitely open to iterating if anyone wants to open a new issue to discuss the second API request.

PR #213, opened 2023-06-08
Add primary provider indicator badge

iandunn:
See #194 

This adds a badge to indicate which provider is the primary/default. I choose "default" because that seems more intuitive for users. 

I couldn't find any other badges in Figma to reference, so I just tried to make it decent. @jasmussen If you're aware of any, or want to make one, let me know.

* Only one ordinary provider active - no change because an indicator isn't needed
    Screenshot 2023-06-08 at 2 53 27 PM
* Two ordinary providers active - badge shows to indicate default
    Screenshot 2023-06-08 at 3 29 05 PM

    Screenshot 2023-06-08 at 3 29 44 PM

### DIFF
--- a/settings/rest-api.php
+++ b/settings/rest-api.php
@@ -165,6 +165,22 @@ function register_user_fields(): void {
 
 	register_rest_field(
 		'user',
+		'2fa_primary_provider',
+		[
+			'get_callback' => function( $user ) {
+				$provider = Two_Factor_Core::get_primary_provider_for_user( get_userdata( $user['id'] ) );
+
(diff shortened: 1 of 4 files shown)


jasmussen:
Can the default be changed? I.e. does the badge need to be interactive?

If it's inert, I think we can just go with a simple text change:

checklist

Otherwise you could have a separate dropdown, like so:
checklist 2

iandunn:
> Can the default be changed? I.e. does the badge need to be interactive?

No, we automatically set WebAuthn as the default if they have both, since it's more secure and often more convenient.

> If it's inert, I think we can just go with a simple text change:

👍🏻

PR #215, opened 2023-06-08
File structure and components refactoring and test update

### LINKED ISSUE #210: Refactor utilities, tests out of components folder
In https://github.com/WordPress/wporg-two-factor/pull/200#discussion_r1219694337 we talked about 

- [x] Creating a new `utilities` folder
- [ ] Rename the original `utilities.js` to `index.js` / `common.js` to indicate its generic use
- [ ] Move the customized hook `useUser` currently in utilities to a new directory called `hooks/`, placed in a file called `useUser.js`
- [ ] Move `tests/` folder outside of `components/`

outdoor2kode:
Fixes #210 

This PR refactors the items mentioned in #210 as well as the following ones

1. Add `renderHook` for useUser test. 

In testing, using `renderHook` is recommended as it simulates a true React environment for Hooks. While some Hooks like `useSelect` and `useEntityRecord` may work outside a true React environment because their functions aren't tied to the React lifecycle, it's not guaranteed. For lifecycle-dependent Hooks like `useEffect`, a real React environment is necessary. Therefore, always use tools like r`enderHook` in tests for accurate simulations, even if some tests pass without them.

This is an issue discovered in https://github.com/WordPress/wporg-two-factor/pull/205.

2. Extract ScreenNavigation for better readability and improved maintainability.
3. Add navigateToScreen as dependency

This is an issue discovered in https://github.com/WordPress/wporg-two-factor/pull/205.

4. Bump version of `@wordpress/scripts` to `25.5.1`

`3.` makes me wonder why there's no warning/error for not adding required dependencies to a hook, then I found that the rule was added later in https://github.com/WordPress/gutenberg/pull/24914 (In this PR, there are some discussions about whether to enable the rule). Since our version of `@wordpress/scripts` is pinned to `24`, it can't reflect the update on Gutenberg side. The latest version of `@wordpress/scripts` is `26.6.0` at the moment, but it wasn't upgraded to that version in this PR as there'd be failures when doing js testing. The root cause of it might come from the older version of `@wordpress/core-data`. At the moment, I guess it's not absolutely necessary to bump the version to `26.6.0`, as the rule I really want to introduce is `'react-hooks/exhaustive-deps': 'warn',` which still exists in the latest version. We could address it when it really needs to.


iandunn (approved):
LGTM 👍🏻

PR #217, opened 2023-06-08
Fix unable to see the first set of backup codes after enabling the 2FA

### LINKED ISSUE #216: Unable to see the first set of backup codes after completing the 2FA application
After completing the 2FA application and entering the backup codes screen, the user is taken to the Manage screen the moment the codes are generated, making the user unable to see the current codes.

https://github.com/WordPress/wporg-two-factor/assets/18050944/28bb0f27-6f2b-482a-ad4e-18e9164387da

outdoor2kode:
Fixes #216, and the behavior of #190 is changed.

This PR adds `isSetupFinished` to `record`. 
It could fix the issue - Unable to see the first set of backup codes after enabling the 2FA,
and also somewhat change the problem raised in https://github.com/WordPress/wporg-two-factor/issues/190, as the flow after the fix here is something like 

* If users don't click the `All finish` button and go back to account screen -> the backup code button status _is_ updated as enabled and clickable (This wasn't fixed by this PR, and actually this is related to the issue of this PR - **the root cause for "as soon as the codes are generated, the screen gets directed to `Manage` screen" is that the `backupCodesEnabled` [here](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/backup-codes.js#L24) has already been set to `true`, and this makes the issue in #190 go away.** i.e., the backup code button status is updated as enabled and clickable) 
However, even though the issue from #190 is resolved, another similar problem has popped up. The status on the account-status screen should _not_ be set to enabled. This is because, after the fix implemented in this PR, there might be some confusion. For instance, (1) if users don't click `All finish` and directly navigate back to the account screen, they would see that the backup codes are enabled. This could be misleading. Similarly, (2) if users have already activated their backup codes once, they might out of curiosity or by accident click the generate new codes button. At this point, since they don't really intend to generate new codes, they might just click `Back` to leave the screen, or even log out, thinking they could still use the backup codes that they activated earlier - which they couldn't. Therefore, the status _shouldn't_ be set to `enabled`, but at the same time, it _shouldn't_ prevent users from clicking it as well. We could handle this in another PR.

* If users click the `All finish` button and go back to account screen ->  the status _is_ correctly updated (enabled and clickable)

For more details about flow please see the screencast below.

This workaround may not seem ideal, but other methods like adding state or using the apiFetch middleware don't really handle all the cases quite well. Once the [upstream PR](https://github.com/WordPress/two-factor/issues/507) is fixed, this can be removed. At that point, we will have the ability to decide when `backupCodesEnabled` should be enabled and control the state.

## Screencast

https://github.com/WordPress/wporg-two-factor/assets/18050944/bc3378ba-30fe-44fe-acb2-fd312cc8d6a0

## Testing Step

1. Set up 2FA while sandboxed.
2. When 2FA is successfully activated, see if it redirects to backup codes screen and stays there, if so, this issue is fixed.

### DIFF
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -33,6 +33,10 @@ export default function BackupCodes() {
 		return;
 	}
 
+	// TODO: record.isSetupFinished and its related logic should be removed
+	// once https://github.com/WordPress/two-factor/issues/507 is fixed.
+	// This is a workaround that fixes the side effect brought up by #507.
+	// See more in https://github.com/WordPress/wporg-two-factor/issues/216 and its PR.
 	if ( backupCodesEnabled && record.isSetupFinished && ! regenerating ) {
 		return <Manage setRegenerating={ setRegenerating } />;
 	}
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -17,7 +17,11 @@ import { refreshRecord } from '../utilities/common';
  */
 export default function BackupCodes() {
 	const {
-		user: { backupCodesEnabled, totpEnabled },
+		user: {
+			backupCodesEnabled,
+			totpEnabled,
+			userRecord: { record },
+		},
 		navigateToScreen,
 	} = useContext( GlobalContext );
 	const [ regenerating, setRegenerating ] = useState( false );
@@ -29,7 +33,7 @@ export default function BackupCodes() {
 		return;
 	}
 
-	if ( backupCodesEnabled && ! regenerating ) {
+	if ( backupCodesEnabled && record.isSetupFinished && ! regenerating ) {
 		return <Manage setRegenerating={ setRegenerating } />;
 	}
 
@@ -81,12 +85,12 @@ function Setup( { setRegenerating } ) {
 
 	// Finish the setup process.
 	const handleFinished = useCallback( async () => {
-		// TODO: Add try catch here after https://github.com/WordPress/wporg-two-factor/pull/187/files is merged.
-		// The codes have already been saved to usermeta, see `generateCodes()` above.
-		await refreshRecord( userRecord ); // This has the intended side-effect of redirecting to the Manage screen.
 		setGlobalNotice( 'Backup codes have been enabled.' );
 		setRegenerating( false );
-	} );
+		userRecord.record.isSetupFinished = true;
+		// The codes have already been saved to usermeta, see `generateCodes()` above.
+		await refreshRecord( userRecord ); // This has the intended side-effect of redirecting to the Manage screen.
+	}, [] );
 
 	return (
 		<>
@@ -171,11 +175,9 @@ function CodeList( { codes } ) {
  */
 function Manage( { setRegenerating } ) {
 	const {
-		user: {
-			userRecord: { record },
-		},
+		user: { userRecord },
 	} = useContext( GlobalContext );
-	const remaining = record[ '2fa_backup_codes_remaining' ];
+	const remaining = userRecord.record[ '2fa_backup_codes_remaining' ];
 
 	return (
 		<>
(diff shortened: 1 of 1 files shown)


settings/src/components/backup-codes.js
```diff
@@ -71,12 +74,13 @@ function Setup( { setRegenerating } ) {
...
-	const handleFinished = useCallback( () => {
-		// The codes have already been saved to usermeta, see `generateCodes()` above.
-		refreshRecord( userRecord ); // This has the intended side-effect of redirecting to the Manage screen.
+	const handleFinished = useCallback( async () => {
 		setGlobalNotice( 'Backup codes have been enabled.' );
 		setRegenerating( false );
-	} );
+		userRecord.record.isSetupFinished = true;
+		// The codes have already been saved to usermeta, see `generateCodes()` above.
+		await refreshRecord( userRecord ); // This has the intended side-effect of redirecting to the Manage screen.
```
outdoor2kode, line 96: Not so sure about the meaning of the comment in this line, does it make sense to remove it now?

StevenDufresne:
> accident click the generate new codes button

Accidentally clicking a button, provided it's a standard button, isn't something to worry about. I think anytime that button is clicked, it's fair and expected that new codes are created. I don't think we need to complicate things by having extra confirmations provided we have recourse for unblocking locked accounts (which we do). 

I would argue that we can always show the backup codes on the code view (we have them saved right?). I'm not quite sure why we treat them as temporary and never show them again?

outdoor2kode:
> I would argue that we can always show the backup codes on the code view (we have them saved right?). I'm not quite sure why we treat them as temporary and never show them again?

I'm not so sure about how it was decided as is in the first place, and I couldn't easily find a best practice for this argument. So I asked chatGPT about it, and here's what it said, it seems to make sense:

> Two-factor authentication (2FA) backup codes are typically shown to the user only once, at the time they are generated. The idea is that the user should make a copy of these codes and keep them in a safe place, separate from their primary device. This is a security measure to prevent unauthorized access to the backup codes.
> 
> If the backup codes were always visible in the user interface, then anyone who managed to gain access to your account could also see and use these backup codes. This would defeat the purpose of having a second factor of authentication.
>
> Therefore, it's standard practice to treat backup codes as temporary and only show them once. If the user loses their backup codes, they should have an option to generate a new set, which will invalidate the old ones. This way, if the old codes were discovered by someone else, they could not be used to access the account.

That being said, I can see that Google always shows the backup codes like you mentioned and put the re-generate button on the same page. But I also can remember there are some services hiding the codes right away after users leave the page. 

image

> Accidentally clicking a button, provided it's a standard button, isn't something to worry about. I think anytime that button is clicked, it's fair and expected that new codes are created. I don't think we need to complicate things by having extra confirmations provided we have recourse for unblocking locked accounts (which we do).

This was brought up [here](https://github.com/WordPress/two-factor/issues/507#issuecomment-1405214682), and it was confirmed as an edge case. If we're not changing to always show the backup codes on the code view, I assume it's still a good-to-have feature, but with low priority, though.

outdoor2kode:
Any further concerns about not merging this PR? If not, I'll squash merge it first and we can come back to revert it anytime we'd like to. Changes here are rather simple, only `isSetupFinished` is added, and there's a clear TODO comment about it.

StevenDufresne:
> If the backup codes were always visible in the user interface, then anyone who managed to gain access to your account could also see and use these backup codes. This would defeat the purpose of having a second factor of authentication

I think this is mostly incorrect, if a bad actor can access the backup codes, they can also access the regenerate button which kills the backup codes and gives them new ones and since we don't have logging you wouldn't know. That is a bigger topic thought.

Also, Github also shows them provided you re-auth.

At any rate, I'm okay with this to merge in the meantime.

iandunn:
In our case at least, it's not possible to show the backup codes after the initial generation, because [they're hashed before being stored in the database](https://github.com/WordPress/two-factor/blob/f21ed3ccc5891735e6563315d62c45638d223874/providers/class-two-factor-backup-codes.php#L242-L249), similar to passwords.

PR #227, opened 2023-08-22
Add success animation to security key and TOTP flows

### LINKED ISSUE #226: Add register keys success animation
See [this comment](https://github.com/WordPress/wporg-two-factor/pull/95#issuecomment-1563945641)

> I imagine it could be a fun little CSS component to animate, and it could be used when the security key is detected, then sit there for 2 seconds, and then automatically redirect you. Potentially we could also add it to similar flows, like successfully connecting 2FA and the likes.

And [this comment](https://github.com/WordPress/wporg-two-factor/issues/194#issuecomment-1672772958) regarding whether the animation can be used throughout the UI (eg. adding TOTP)

> If the actions are small and not too substantial, I think a notice/snackbar can be fine.

> In the flow suggested above, the animation exists because the action is a bit more substantial, i.e. connecting a security key. I imagine scanning a QR code is also a substantial action. Deleting a key, or something like that, would not be substantial and can just be confirmed more simply.

adamwoodnz:
Closes #226 

- Clears notices from other actions in WebAuthn UI
- Adds a success animation after a key is registered or TOTP is enabled, see issue for specs and [comment](https://github.com/WordPress/wporg-two-factor/issues/194#issuecomment-1672772958) regarding using the animation in places other than WebAuthn.

webauthn success

totp success

## Testing

Checkout this branch in your sandbox.

WebAuthn still has a development only flag, so either do a dev build of `settings` with `NODE_ENV=development wp-scripts build --webpack-copy-php`, or a prod build and remove the [dev flag](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/account-status.js#L60).

For TOTP, disable if enabled, then re-enable. You should see the success message and animation after entering the 6 digit code to enable your app.

For WebAuthn, register a new key and you should see the success message and animation when the key has been successfully added.

You should also test deleting a key so that the delete success notice is displayed. When you enter the register key flow, that notice should be dismissed.

### DIFF
--- a/settings/src/components/success.js
+++ b/settings/src/components/success.js
@@ -1,0 +1,36 @@
+/**
+ * WordPress dependencies
+ */
+import { useState } from '@wordpress/element';
+import { Icon, check } from '@wordpress/icons';
+
+/**
+ * Render the "Success" component.
+ *
+ * Shows a message and animation, then calls the `afterTimeout` callback
+ *
+ * @param props
+ * @param props.afterTimeout
+ * @param props.message
+ */
+export default function Success( { message, afterTimeout } ) {
+	const [ hasTimer, setHasTimer ] = useState( false );
+
+	if ( ! hasTimer ) {
+		// Time matches the length of the CSS animation property on .wporg-2fa__success
+		setTimeout( afterTimeout, 4000 );
+		setHasTimer( true );
+	}
+
+	return (
+		<>
+			<p className="wporg-2fa__screen-intro">{ message }</p>
+
+			<p className="wporg-2fa__webauthn-register-key-status" aria-hidden>
+				<div className="wporg-2fa__success">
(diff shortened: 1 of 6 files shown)


iandunn (approved):
LGTM and works in my tests 👍🏻

adamwoodnz:
@jasmussen this has shipped and you can test the animation by (re)enabling your 2FA app. Please let me know if you'd like to tweak it 🙂

jasmussen:
Cool! 

I'd love to tweak two things:

1. Add a min-height to the white container box so it doesn't collapse when the animation appears, and ideally center that animation in the available whitespace that makes.
2. The start of the animatin is a bit abrupt, can it start slightly delayed? It's okay that this animation takes a while, because it is not UI blocking, and it's celebrating an important security advancement you took.

PR #230, opened 2023-08-24
Update animations and modals

### LINKED ISSUE #225: Add confirmation dialog before disabling TOTP
We have a confirmation dialog before disabling security keys. For consistency it feels like this important action should also follow this pattern.

### LINKED ISSUE #229: Update animations and modals
See @jasmussen's [comment](https://github.com/WordPress/wporg-two-factor/pull/227#issuecomment-1689454961)

> Add a min-height to the white container box so it doesn't collapse when the animation appears, and ideally center that animation in the available whitespace that makes.

> The start of the animatin is a bit abrupt, can it start slightly delayed? It's okay that this animation takes a while, because it is not UI blocking, and it's celebrating an important security advancement you took.

The action buttons in the WebAuthn modals should also be hidden while actions are processing; there is no way to cancel once in progress, and triggering again will result in errors.

adamwoodnz:
Closes #229, #225

- Adds a minimum height to the TOTP and Security Key success animations
- Adds a slight delay to the start of the success animation
- Hides the action buttons in the Security Key modals when actions are processing
- Adds a modal to confirm disabling TOTP, to be consistent with security keys actions (#225)

### Disable security keys

(Error is expected)

### DIFF
--- a/settings/src/components/account-status.scss
+++ b/settings/src/components/account-status.scss
@@ -35,6 +35,7 @@
 			.wporg-2fa__status-icon {
 				grid-area: status;
 				align-self: center;
+				margin: 0;
 			}
 
 			h3 {
--- a/settings/src/components/account-status.scss
+++ b/settings/src/components/account-status.scss
@@ -35,7 +35,6 @@
 			.wporg-2fa__status-icon {
 				grid-area: status;
 				align-self: center;
-				margin: 0;
 			}
 
 			h3 {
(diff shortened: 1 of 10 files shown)


settings/src/components/account-status.scss
```diff
@@ -35,6 +35,7 @@
 			.wporg-2fa__status-icon {
 				grid-area: status;
 				align-self: center;
+				margin: 0;
```
adamwoodnz, line 38: This is now a common classname, and in this instance it doesn't require margin

iandunn (approved):
LGTM 👍🏻

adamwoodnz:
This has shipped @jasmussen, hoping everything should be smoother and more consistent now. Let me know if it needs further attention.

jasmussen:
I won't have time to easily test this, but there's always room for Q/A and I'll be sure to raise an issue if I stumble on this in the future!

PR #233, opened 2023-09-04
Improve disable keys flow

### LINKED ISSUE #232: Disabling WebAuthn doesn't close modal
1. Switch to the `disable_provider` branch of Two Factor (https://github.com/WordPress/two-factor/pull/587)
1. Register a WebAuthn key and enable the provider
3. Click `Disable Security Keys`

The provider will be disabled, but the modal isn't closed:

Screenshot 2023-09-01 at 4 11 26 PM

adamwoodnz:
Closes #232 

- Close the confirmation modal on both success and failure
- Hide the action buttons when enabling
- Improve the messaging on the accounts screen when disabled
- Clear success notices when toggling enablement

### DIFF
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -65,7 +65,7 @@ export default function AccountStatus() {
 					bodyText={
 						webAuthnEnabled
 							? 'You have two-factor authentication enabled using security keys.'
-							: 'You have not registered any security keys.'
+							: 'You have not enabled security keys for two-factor authentication.'
 					}
 					isPrimary={ 'TwoFactor_Provider_WebAuthn' === primaryProvider && totpEnabled }
 				/>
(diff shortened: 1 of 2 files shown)


settings/src/components/account-status.js
```diff
@@ -65,7 +65,7 @@ export default function AccountStatus() {
 					bodyText={
 						webAuthnEnabled
 							? 'You have two-factor authentication enabled using security keys.'
-							: 'You have not registered any security keys.'
+							: 'You have not enabled security keys for two-factor authentication.'
```
adamwoodnz, line 68: You can have security keys registered but not have the provider enabled, so this is more accurate, and also more consistent with the TOTP message [below](https://github.com/WordPress/wporg-two-factor/pull/233/files#diff-8a81bc73a741dacb0ff448a75ab1f1fa1809145d97c1f89b6706a4e2da0c4a9fR81).

iandunn (approved):
LGTM 👍🏻

PR #234, opened 2023-09-11
Update the screen variable ASAP.

### LINKED ISSUE #222: I cannot find the options to activate
I read that it was this link https://wordpress.org/support/users/username/edit/account

But I can't find anything to change it. I can edit the forum profile and the .org profile but I don't see anything about the two factor.

StevenDufresne:
Fixes #222 

This PR fixes a broken screen in production where users visits `https://wordpress.org/support/users/{username}/edit/account`. I was unable to reproduce this locally, so this does require a sandbox. My guess is that the local environment doesn't have as much latency as production and rendering happens more quickly. I have not verified that assumption.

## What's happening?
I've described the issue in [this comment](https://github.com/WordPress/wporg-two-factor/issues/222#issuecomment-1704458307). Essentially we have some default logic where we use the URL to determine the current screen on load. That doesn't work properly.

## Fix

We can fix it by using the `setScreen` function and setting the screen immediately. There are probably some refactors that would make this more predictable but this will work in the interim.

## How to test
1. Before applying this patch, reproduce issue:
   - Log into your wp account
   - Navigate to https://wordpress.org/support/users/{username}/edit/account/
     - Notice the error in your console
   -  Navigate to https://wordpress.org/support/users/{username}/edit/account/?screen=account-status
      - Notice it works
2. Copy your `settings/build/script.js` into your sandbox build folder
3. Navigate to https://wordpress.org/support/users/{username}/edit/account/
     - It should load properly

### DIFF
--- a/settings/src/script.js
+++ b/settings/src/script.js
@@ -63,7 +63,7 @@ function Main( { userId } ) {
 	const [ error, setError ] = useState( '' );
 
 	let currentUrl = new URL( document.location.href );
-	let initialScreen = currentUrl.searchParams.get( 'screen' );
+	const initialScreen = currentUrl.searchParams.get( 'screen' );
 	const [ screen, setScreen ] = useState( initialScreen );
 
 	// The index is the URL slug and the value is the React component.
@@ -83,11 +83,8 @@ function Main( { userId } ) {
 	// The screens where a recent two factor challenge is required.
 	const twoFactorRequiredScreens = [ 'webauthn', 'totp', 'backup-codes' ];
 
-	if ( ! components[ initialScreen ] ) {
-		initialScreen = 'account-status';
-		setScreen( initialScreen );
-		currentUrl.searchParams.set( 'screen', initialScreen );
-		window.history.pushState( {}, '', currentUrl );
+	if ( ! components[ screen ] ) {
+		setScreen( 'account-status' );
 	}
 
 	// Listen for back/forward button clicks.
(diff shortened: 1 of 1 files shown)


outdoor2kode:
> I was unable to reproduce this locally, so this does require a sandbox

I can see the error messages locally with the same testing steps.

I pushed a commit that aligns with the React practice a bit. Tested locally and it works well.

> There are probably some refactors that would make this more predictable but this will work in the interim.

`currentUrl` could also benefit from some refactoring. There's also a hook that is related to it being used in condition. But this might need some work.

settings/src/script.js
```diff
@@ -83,11 +83,8 @@ function Main( { userId } ) {
 	// The screens where a recent two factor challenge is required.
 	const twoFactorRequiredScreens = [ 'webauthn', 'totp', 'backup-codes' ];
 
-	if ( ! components[ initialScreen ] ) {
-		initialScreen = 'account-status';
-		setScreen( initialScreen );
-		currentUrl.searchParams.set( 'screen', initialScreen );
-		window.history.pushState( {}, '', currentUrl );
+	if ( ! components[ screen ] ) {
+		setScreen( 'account-status' );
```
outdoor2kode, line 87: I initially wanted to extract this logic. Just tried out a few methods, there's still one error message remaining. But the original issue has disappeared.

StevenDufresne:
> I pushed a commit that aligns with the React practice a bit. Tested locally and it works well.

Thanks. Anything left needed before merge?

outdoor2kode (approved):
> Thanks. Anything left needed before merge?

No on my side 👍

PR #235, opened 2023-09-11
Remove the development flags.

StevenDufresne:
This removes the developer flags to turn on WebAuthn.

PR #236, opened 2023-09-11
Try moving security keys below the one time passwords via auth app.

StevenDufresne:
This PR builds on #235 and moves the WebAuthN component down below the TOTP control. I don't know if this is the best idea but I thought a PR would be a good place to discuss (maybe it was already discussed?).

## Why
- We don't have a clear path to configure things properly yet, so top to bottom is likely the most logical
- TOTP works from multiple devices and we probably want it configured as a backup plan for users.
  - Maybe this is a dated idea though. 

| Current | After |
| --- | --- |
| Screenshot 2023-09-11 at 2 57 10 PM | Screenshot 2023-09-11 at 2 57 26 PM|

### DIFF
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -56,20 +56,17 @@ export default function AccountStatus() {
 				}
 			/>
 
-			{ /* TODO: Only enable WebAuthn UI in development, until it's finished. */ }
-			{ 'development' === process.env.NODE_ENV && (
-				<SettingStatusCard
-					screen="webauthn"
-					status={ hasPrimaryProvider && ! webAuthnEnabled ? 'info' : webAuthnEnabled }
-					headerText="Two-Factor Security Key"
-					bodyText={
-						webAuthnEnabled
-							? 'You have two-factor authentication enabled using security keys.'
-							: 'You have not enabled security keys for two-factor authentication.'
-					}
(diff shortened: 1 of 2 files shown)


StevenDufresne:
I found another test case, where we need TOTP installed. Revalidation. I am unable to go through the revalidation sequence once my totp is disabled even with security keys enabled.

Screenshot 2023-09-11 at 3 26 02 PM

You need to log out and log back in to fix this.

iandunn:
🤔 The revalidation issue sounds like a bug. I don't think the intention was to require TOTP if WebAuthn is active. Some folks may prefer to only have WebAuthn and backup codes, and that seems reasonable to me.

I'd personally keep WebAuthn first, since it's a much stronger mechanism IMO. There's some precedent for that in: https://github.com/WordPress/wporg-two-factor/blob/8dce71504d86f8204b9528d4214d803f08eefee9/wporg-two-factor.php#L140-L143

If we fix the revalidation bug, is there still a reason to move it below TOTP? WebAuthn also works from multiple devices with hardware keys. With passkeys, one can be registered for each device.

I don't feel strongly either way, though.

StevenDufresne:
Fair enough. I'm ok with that. I've tested the potential regression, seems like it's my local environment.

PR #237, opened 2023-09-12
Update Remove security key modal

StevenDufresne:
The delete key modal currently widens significantly based on the name of the key. It also is too verbose as we mention the key name in the title, content and remove button. Lastly, the originating interface uses "delete" language but the modal uses remove. 

**Changes:**
- Remove the key name from title and button
  - Since we no longer have the key name in the button, adds `isDestructive` to make it red
- Changes "Remove" wording to "Delete" to make it align with the underlying language.

| Before | After |
|--------|--------|
| Screenshot 2023-09-12 at 9 57 43 AM | Screenshot 2023-09-12 at 9 54 48 AM |

### DIFF
--- a/settings/src/components/webauthn/list-keys.js
+++ b/settings/src/components/webauthn/list-keys.js
@@ -95,7 +95,7 @@ export default function ListKeys() {
 function ConfirmRemoveKey( { keyToRemove, onConfirm, onClose, deleting, error } ) {
 	return (
 		<Modal
-			title={ `Delete security key` }
+			title="Delete security key"
 			className="wporg-2fa__confirm-delete-key"
 			onRequestClose={ onClose }
 		>
--- a/settings/src/components/webauthn/list-keys.js
+++ b/settings/src/components/webauthn/list-keys.js
@@ -3,7 +3,7 @@
  */
 import { Button, Modal, Notice, Spinner } from '@wordpress/components';
 import { useCallback, useContext, useState } from '@wordpress/element';
-import { Icon, cancelCircleFilled, key } from '@wordpress/icons';
+import { Icon, cancelCircleFilled, key as keyIcon } from '@wordpress/icons';
 
 /**
  * Internal dependencies
@@ -102,7 +102,7 @@ function ConfirmRemoveKey( { keyToRemove, onConfirm, onClose, deleting, error } 
 			<p className="wporg-2fa__screen-intro">
 				Are you sure you want to delete the following key?
 				<span className="wporg-2fa__screen-key">
-					<Icon icon={ key } />
+					<Icon icon={ keyIcon } />
 					<span>{ keyToRemove.name }</span>
 				</span>
 			</p>
--- a/settings/src/components/webauthn/list-keys.js
+++ b/settings/src/components/webauthn/list-keys.js
@@ -95,12 +95,15 @@ export default function ListKeys() {
 function ConfirmRemoveKey( { keyToRemove, onConfirm, onClose, deleting, error } ) {
 	return (
 		<Modal
-			title={ `Remove ${ keyToRemove.name }?` }
+			title={ `Delete security key` }
 			className="wporg-2fa__confirm-delete-key"
 			onRequestClose={ onClose }
 		>
 			<p className="wporg-2fa__screen-intro">
-				Are you sure you want to remove the <code>{ keyToRemove.name }</code> security key?
+				Are you sure you want to delete the following key?
+				<ul className="wporg-2fa__screen-list">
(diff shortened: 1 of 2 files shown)


settings/src/components/webauthn/list-keys.js
```diff
@@ -95,12 +95,15 @@ export default function ListKeys() {
...
 		<Modal
-			title={ `Remove ${ keyToRemove.name }?` }
+			title="Delete security key"
 			className="wporg-2fa__confirm-delete-key"
 			onRequestClose={ onClose }
 		>
 			<p className="wporg-2fa__screen-intro">
-				Are you sure you want to remove the <code>{ keyToRemove.name }</code> security key?
+				Are you sure you want to delete the following key?
+				<ul className="wporg-2fa__screen-list">
```
adamwoodnz: Why a list? I get that you're after a bulleted style, but semantically it doesn't make sense to me if there's only ever going to be one item. Maybe a key icon or some other treatment could work?

settings/src/components/webauthn/list-keys.js
```diff
@@ -109,8 +112,8 @@ function ConfirmRemoveKey( { keyToRemove, onConfirm, onClose, deleting, error }
 				</div>
 			) : (
 				<div className="wporg-2fa__submit-actions">
-					<Button variant="primary" onClick={ onConfirm }>
-						Remove { keyToRemove.name }
+					<Button variant="primary" isDestructive onClick={ onConfirm }>
```
adamwoodnz, line 116: I like the message this treatment conveys. There are other modal buttons which could use the same; disable TOTP and disable security keys

settings/src/components/webauthn/list-keys.js
StevenDufresne: Yeah, understood. It was the easiest approach. 

There's a small chance that this could be used for deleting multiple keys in the future but that definitely isn't set in stone. I can change it to something else if it doesn't make sense.

settings/src/components/webauthn/list-keys.js
StevenDufresne: I think it's useful for things that aren't reversible, like deleting keys. For disabling, I can see it go both ways since you can re-enable them afterwards. 

But since you can non-destructively turn off keys, but destructively disable the TOTP (because you have to set it up again), maybe a blanket "red" button does makes sense.

settings/src/components/webauthn/list-keys.js
```diff
@@ -109,8 +112,8 @@ function ConfirmRemoveKey( { keyToRemove, onConfirm, onClose, deleting, error }
...
+						Delete
```
iandunn, line 117: IIRC, repeating the name was intentional, to make sure the user is aware _which_ key they're deleting. It'd be easy to click on a different key and then accidentally delete the wrong one. Similar to how github makes you type out the name of the repository when it confirms you want to delete it, but not that extreme.

I don't personally see the harm in repeating it, or the modal changing size. An alternative would be wrapping the title in a `<code>`, quotation marks, etc to distinguish it from the regular text.

I don't feel strongly, though.

iandunn (approved):
`isDestructive` is a really good addition 👍🏻

settings/src/components/webauthn/list-keys.js
StevenDufresne: I understand the intention and think it can be effective. I'm not certain it's needed here, although it's probably too early to be completely certain of that. This action is undoable but isn't as significant as deleting a repository. It's more a kin to deleting an ssh key, and GitHub itself doesn't even reference the key in that window (although they use very strong language).

> I don't personally see the harm in repeating it
I don't know if its about harm, but more about readability. The repetition make me more likely to skip most of the content because I scan for the recurring pattern and skip past it. 
It can also look somewhat broken depending on what the the user named the key. 

I'll get a second opinion.

settings/src/components/webauthn/list-keys.js
StevenDufresne: The key icon is pretty nice.

Screenshot 2023-09-13 at 2 41 06 PM

settings/src/components/webauthn/list-keys.js
adamwoodnz: IMO [this version](https://github.com/WordPress/wporg-two-factor/pull/237#discussion_r1323996058) is more pleasing.

I wonder whether bolding the name of the key might bring more attention to it, if we're concerned about ensuring the correct key.

iandunn: Sounds good to me 👍🏻

PR #239, opened 2023-09-13
 Show error message when server rejects new password

### LINKED ISSUE #53: Handle errors in `savePasswordHandler` response

iandunn:
Fixes #53 

Screenshot 2023-09-13 at 9 40 14 AM

### Testing

Add something like this to an mu-plugin, then try to save a new pw.

```php
function validate_rest_password( $result, WP_REST_Server $server, WP_REST_Request $request ) {
	if ( is_wp_error( $result ) ) {
		return $result;
	}

	if ( ! str_starts_with( $request->get_route(), '/wp/v2/users' ) ) {
		return $result;
	}

	if ( false === stripos( WP_REST_SERVER::EDITABLE, $request->get_method() ) ) {
		return $result;
	}

	$params = $request->get_params();

	if ( ! isset( $params['id'], $params['password'] ) ) {
		return $result;
	}

	$user = get_user_by( 'id', $params['id'] );

	if ( ! $user ) {
		return $request;
	}

	$result = new WP_Error( 'rest_password_weak', __( 'Password is too weak.' ), array( 'status' => 400 ) );

	return $result;
}
add_filter( 'rest_pre_dispatch', __NAMESPACE__ . '\validate_rest_password', 1, 3 );
```

### Misc

I also added some missing hook dependencies that eslint was complaining about. The commits are atomic, so you can review that separately if you want.

### DIFF
--- a/settings/src/components/password.js
+++ b/settings/src/components/password.js
@@ -47,7 +47,7 @@ export default function Password() {
 		}
 
 		setGlobalNotice( '' );
-	}, [ hasEdits ] );
+	}, [ hasEdits, setGlobalNotice ] );
 
 	/**
 	 * Handle clicking the `Generate Password` button.
@@ -55,7 +55,7 @@ export default function Password() {
 	const handlePasswordGenerate = useCallback( async () => {
 		edit( { password: generatePassword( 24, true, true ) } );
 		setInputType( 'text' );
-	}, [] );
+	}, [ edit ] );
 
 	// Handle form submission.
 	const handleFormSubmit = useCallback(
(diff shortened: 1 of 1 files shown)


StevenDufresne (approved):
This worked using the test case. The incidental linter fixes don't seem to introduce any issues either.

👍

PR #245, opened 2023-09-14
Prevent unintentional backup code regeneration

### LINKED ISSUE #244: Backup Codes unintentionally regenerated when switching devices, clearing local storage, etc
1. Setup Backup Codes
2. Delete the `WPORG_2FA_HAS_BACKUP_CODES_BEEN_SAVED` local storage item
3. Refresh

Your original Backup Codes from step `1` will be regenerated. Step `2` artificially creates the situation, but the same thing should happen when viewing the screen from your phone when it was originally setup on your laptop, switching to a different browser,  clearing your browser data, etc.

That was mentioned in #221, but I'm not sure the implications were obvious. When this happens it means that the codes that folks originally saved (maybe months or years prior) will be overwritten, and they'll have to save the new ones, or they risk being locked out of their account. At best that's annoying/inconvenient, but I worry that it'll create situations where folks didn't write down the new keys and lose their account.

iandunn:
Fixes #244

This removes the local storage item, and instead uses the number of remaining backup codes to determine if the provider has already been setup. That comes from the database, so it spans browsers/devices/sessions.

I wasn't able to reproduce the bug mentioned in #221, so I don't think this will cause a regression, but I'd appreciate some more eyes to double check.

I initially ran into the bug from #217, but was able to fix that by refreshing the user when navigating to Account Status.

### DIFF
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -181,11 +181,8 @@ function CodeList( { codes } ) {
  */
 function Manage( { setRegenerating } ) {
 	const {
-		user: {
-			userRecord: { record },
-		},
+		user: { backupCodesRemaining },
 	} = useContext( GlobalContext );
-	const remaining = record[ '2fa_backup_codes_remaining' ];
 
 	return (
 		<>
@@ -196,18 +193,19 @@ function Manage( { setRegenerating } ) {
 					code can only be used once.
 				</p>
 
-				{ remaining > 5 && (
+				{ backupCodesRemaining > 5 && (
 					<p>
-						You have <strong>{ remaining }</strong> backup codes remaining.
+						You have <strong>{ backupCodesRemaining }</strong> backup codes remaining.
 					</p>
 				) }
 
-				{ remaining <= 5 && (
+				{ backupCodesRemaining <= 5 && (
 					<Notice status="warning" isDismissible={ false }>
 						<Icon icon={ warning } />
-						You only have <strong>{ remaining }</strong> backup codes remaining. Please
-						regenerate and save new ones before you run out. If you don&apos;t, you
-						won&apos;t be able to log into your account if you lose your phone.
+						You only have <strong>{ backupCodesRemaining }</strong> backup codes
+						remaining. Please regenerate and save new ones before you run out. If you
+						don&apos;t, you won&apos;t be able to log into your account if you lose your
+						phone.
 					</Notice>
 				) }
 			</div>
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -18,16 +18,9 @@ import ScreenLink from './screen-link';
  */
 export default function BackupCodes() {
 	const {
-		user: { backupCodesEnabled, hasPrimaryProvider },
+		user: { backupCodesEnabled, hasPrimaryProvider, backupCodesRemaining },
 	} = useContext( GlobalContext );
 	const [ regenerating, setRegenerating ] = useState( false );
-	// TODO: hasSetupCompleted and its related logic should be removed
-	// once https://github.com/WordPress/two-factor/issues/507 is fixed.
-	// This is a workaround that fixes the side effect brought up by #507.
-	// See more in https://github.com/WordPress/wporg-two-factor/issues/216 and its PR.
-	const [ hasSetupCompleted, setHasSetupCompleted ] = useState(
-		localStorage.getItem( 'WPORG_2FA_HAS_BACKUP_CODES_BEEN_SAVED' ) === 'true'
-	);
 
 	// Prevent users from accessing directly through the URL.
 	if ( ! hasPrimaryProvider ) {
@@ -44,13 +37,11 @@ export default function BackupCodes() {
 		);
 	}
 
-	if ( backupCodesEnabled && hasSetupCompleted && ! regenerating ) {
-		return <Manage setRegenerating={ setRegenerating } />;
+	if ( ! backupCodesEnabled || backupCodesRemaining === 0 || regenerating ) {
+		return <Setup setRegenerating={ setRegenerating } />;
 	}
 
-	return (
-		<Setup setRegenerating={ setRegenerating } setHasSetupCompleted={ setHasSetupCompleted } />
-	);
+	return <Manage setRegenerating={ setRegenerating } />;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


iandunn:
This looks like it regresses the solution in #217, so I'm looking into that.

iandunn:
I think [93df040](https://github.com/WordPress/wporg-two-factor/pull/245/commits/93df040d135e2d1e8cf18b02e71dd131baa2a14f) fixes the #217 regression.

outdoor2kode:
https://github.com/WordPress/wporg-two-factor/assets/18050944/4947cdaf-4ff9-422d-866c-7a9be9590bba

The first half of the video is testing #217 (Unable to see the first set of backup codes after enabling the 2FA). It seems that after completing the activation of 2FA, the initial set of backup codes still can't be seen. 
The second half of the video (starting from switching to the Profile and then switching back) is testing #221 (When leaving the account page and returning to the backup codes page, a new set of backup codes is regenerated without asking). There's no regression in this part, it doesn't automatically generate new backup codes after leaving the account page and returning.

Regarding the issue with #217, it seems that the regression is not from the changes made here. It's already like this in the current production.

**Update**
> I think [93df040](https://github.com/WordPress/wporg-two-factor/pull/245/commits/93df040d135e2d1e8cf18b02e71dd131baa2a14f) fixes the https://github.com/WordPress/wporg-two-factor/pull/217 regression.

I found that this commit didn't seem to handle this issue: `(Unable to see the first set of backup codes after enabling the 2FA)`. What's the issue it's trying to address? Is it making the account screen show the correct status?

iandunn:
🤔 I can't reproduce the problem in the video. This is what I'm doing:

1. `wp user meta list iandunn-test |grep factor`
1. `wp user meta delete iandunn-test _two_factor_backup_codes`. Sometimes I've seen multiple `_two_factor_backup_codes` entries in the database, I think because React's Strict Mode causes the hook to fire twice. So you need to make sure you delete both.
1. `wp user meta delete iandunn-test _two_factor_totp_key`
2. Open https://wordpress.org/support/users/iandunn-test/edit/account/
3. Enable TOTP

When it navigates to the Backup Codes screen, it generates new codes like normal.

Can you check that you don't have any backup codes saved previously? The entry on `AccountStatus` may just be disabled b/c a primary provider isn't enabled, rather than because there aren't any codes saved.

If that isn't it, can you see anything that might be going wrong?

> What's the issue it's trying to address? Is it making the account screen show the correct status?

#244 has a description of the problem and reproduction steps

iandunn:
I rebased to fix the merge conflict, but still can't reproduce the problem you saw.

iandunn:
> > What's the issue it's trying to address? Is it making the account screen show the correct status?

> https://github.com/WordPress/wporg-two-factor/issues/244 has a description of the problem and reproduction steps

Er, nevermind, I realize now that you were asking about 93df040 rather than the overall PR. IIRC, the regression that I saw before 93df040 was that visiting the Backup Codes screen would regenerate new codes every time, rather than recognizing that they had been saved previously and displaying the Manage screen.

outdoor2kode:
> 🤔 I can't reproduce the problem in the video. This is what I'm doing:
> 
> 1. `wp user meta list iandunn-test |grep factor`
> 2. `wp user meta delete iandunn-test _two_factor_backup_codes`. Sometimes I've seen multiple `_two_factor_backup_codes` entries in the database, I think because React's Strict Mode causes the hook to fire twice. So you need to make sure you delete both.
> 3. `wp user meta delete iandunn-test _two_factor_totp_key`
> 4. Open https://wordpress.org/support/users/iandunn-test/edit/account/
> 5. Enable TOTP
> 
> When it navigates to the Backup Codes screen, it generates new codes like normal.

Following your steps worked for me. Before, I always went through Dashboard > Users > Profile to uncheck the `Enabled` option under `Two-Factor Options`. It seems that unchecking `Recovery Codes` there doesn't actually delete the `_two_factor_backup_codes` from the database. 🤔 

>  visiting the Backup Codes screen would regenerate new codes every time, rather than recognizing that they had been saved previously and displaying the Manage screen.

I've figured out why it was initially set up that way:

The original intention was for the backup codes badge in the account status to show as "enabled" only when a user checks "I have printed or saved these codes". If it's not checked, the badge would consistently display as "not enabled", reminding the user that they haven't noted down the backup codes somewhere. But I believe in many actual use cases, users might see the generated backup codes and save them, then directly hit "Back". This would leave them confused as to why the status is "not enabled". If they go into the backup codes screen again, new codes are generated. They might hit "Back" again and again... until they realize they need to check "I have printed or saved these codes" to actually enable it.

So, I think the current process is better: If users note it down and then press "Back", the status shows as "enabled". Or if they don't note it down and press "Back", the status still displays as "enabled", but they can return and generate new codes. 

Given this, I feel the "I have printed or saved these codes" option may no longer be that necessary?

outdoor2kode (approved):
Apart from a small follow-up question and an issue that might be related to an upstream repo, as mentioned in the comment, everything looks good to me. 👍

iandunn:
> Given this, I feel the "I have printed or saved these codes" option may no longer be that necessary?

🤔 I think it's still helpful, to reinforce to the user that it's really important to write them down.

Once https://github.com/WordPress/two-factor/issues/507 if fixed, then we can update our client to not save the codes until the user clicks "All Finished". I created #256 to track that.

StevenDufresne:
I've reverted this PR as it's causing a couple issues:
- It creates some problems with buttons appearing to be pressed before they are listening to the refresh state
- It forces components to re-render too often introducing fatal errors (password view)

Screen Capture on 2023-09-21 at 09-23-53
Screen Capture on 2023-09-21 at 09-14-52

outdoor2kode:
> It creates some problems with buttons appearing to be pressed before they are listening to the refresh state
It forces components to re-render too often introducing fatal errors (password view)

I think they are both reflecting the same situation - when pressing 'Back' to return to the Account Screen, it would first do `refreshRecord` on the current screen and then go back. Since it's using `refreshRecord(userRecord);` [here](https://github.com/WordPress/wporg-two-factor/pull/245/commits/5d99f1c79933ad1e0153831fb97d5f3deef80de0#diff-7eb8cf2231795f70529459a7212a92b4521001f290e37a74596a010b72e02914R46), it tries to save with an empty password on the password screen, triggering `zxcvbn` and causing an error. 

I guess this can be solved by not refreshing all records, but only the backup codes part.

iandunn:
Thanks for catching that! I think #257 will fix it, do you mind checking it out?

PR #247, opened 2023-09-15
Make the disable button destructive.

StevenDufresne:
We introduced the `isDestructive` button prop in #237. This PR adds the same property to the modal that disable keys since it may affect the ability for users to log in. 

Screenshot 2023-09-15 at 11 04 36 AM


adamwoodnz (approved):
LGTM, not doing the same for TOTP? I thought in the previous discussion you felt that TOTP was more destructive.

StevenDufresne:
A dang, right, missed that one. I'll add it.

PR #248, opened 2023-09-15
Add new custom error for aborts.

StevenDufresne:
When a user cancels out of creating a key, `await navigator.credentials.create( { publicKey } )` returns `null`. Currently, we don't check the return and pass it along to the `ajax` which returns a confusing error message. 

This PR improves the error message with one caveat. Since `navigator.credentials.create` is an interface, aborting is potentially not the only reason why the promise will return `null` so we can't assume its from the user aborting. Ideally, we don't show a message on abort but that likely isn't possible. 

| Before | After |
|--------|--------|
| Screenshot 2023-09-15 at 11 48 35 AM | Screenshot 2023-09-15 at 11 48 13 AM |

### DIFF
--- a/settings/src/components/webauthn/register-key.js
+++ b/settings/src/components/webauthn/register-key.js
@@ -50,6 +50,18 @@ export default function RegisterKey( { onSuccess, onCancel } ) {
 
 				const credential = await navigator.credentials.create( { publicKey } );
 
+				/**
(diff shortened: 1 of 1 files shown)


iandunn (approved):
LGTM 👍🏻

outdoor2kode (approved):
LGTM 👍

PR #251, opened 2023-09-15
Enable Backup Codes screen if user has primary provider

### LINKED ISSUE #246: Can't access backup codes screen after they run out
1. Setup backup codes
2. `wp user meta delete {username} _two_factor_backup_codes`

The menu item on the Account Status screen is disabled.

Introduced in 71392546abacecd7a8049ce5ceade566d68fb93b. It seems like maybe it should be based on `hasPrimaryProvider` instead.

Related #243

iandunn:
Fixes #246 

Before this, it couldn't be accessed after the codes are deleted or expire.

### DIFF
--- a/settings/src/components/account-status.js
+++ b/settings/src/components/account-status.js
@@ -88,7 +88,7 @@ export default function AccountStatus() {
 				status={ backupCodesEnabled }
 				headerText="Two-Factor Backup Codes"
 				bodyText={ backupBodyText }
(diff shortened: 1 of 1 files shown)


StevenDufresne (approved):
That makes sense based on the code. I can't speak to the availability of the `Two_Factor_Backup_Codes` provider. I'm assuming it's baked in deep enough that we can assume back up codes will be available.

Issue #243, opened 2023-09-14
Can't create backup keys if WebAuthn enabled and TOTP disabled


iandunn:
1. Wipe out all settings/providers so you start fresh
2. Enable WebAuthn

- [x] You also won't be able to manually access Backup Codes. I think the following code from #189 should be updated to check if the user `hasPrimaryProvider`.
    https://github.com/WordPress/wporg-two-factor/blob/fa4c81b35d0c6b8fe75723ee2529299a0b2e767e/settings/src/components/backup-codes.js#L32-L37

iandunn:
Fixed in #243

PR #252, opened 2023-09-15
Backup Codes: Show error when primary provider not enabled

iandunn:
See #243 

This checks for `hasPrimaryProvider` instead of `totpEnabled`, so that users with WebAuthn instead of TOTP can enable backup codes.

It also shows an error message instead of redirecting, so the user knows what's happening and why. I feel like this is a better UX, but am open to other ideas. It's not essential to the PR.


StevenDufresne (approved):
Tested, looks good. 

I think the redirect works well. 👍

PR #255, opened 2023-09-19
Prevent CSS from loading on unrelated pages

### LINKED ISSUE #35: JS + CSS should not be loaded on every page
As currently expected, the wporg-two-factor settings screen CSS/JS is loaded/output on every page.

```
$ curl -is https://wordpress.org/ | grep wporg-two-factor

<style id='wporg-two-factor-settings-style-inline-css'>
.wp-block-wporg-two-factor-settings h3 {
.wp-block-wporg-two-factor-settings .components-button {
<script src='https://wordpress.org/wp-content/plugins/wporg-two-factor/settings/build/script.js?ver=1672716006' id='wporg-two-factor-settings-script-js'></script>
```

iandunn:
Fixes #35 

The check for `bbp_get_displayed_user_id()` isn't perfect, since it'll still load the stylesheet on a handful of pages that don't display the block. I think it's good enough for a quick workaround, though. Hopefully we can [move to `viewStyle` in the future](https://github.com/WordPress/gutenberg/issues/54491) and remove this code.

To test, look for the `wporg-two-factor-settings-style-css` stylesheet ID in response source, or the `wp-content/plugins/wporg-two-factor/settings/build/style-index.css` URL in the Network tab of browser dev tools.

```shell
> curl -is https://wpms.test/ | grep wporg-two-factor

> curl -is https://wpms.test/users/iandunn/edit/account/ | grep wporg-two-factor
<link rel='stylesheet' id='wporg-two-factor-settings-style-css' href='https://wpms.test/content/plugins/wporg-two-factor/settings/build/style-index.css?ver=0.1.3' type='text/css' media='all' />
```

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -9,6 +9,7 @@ require __DIR__ . '/rest-api.php';
 
 add_action( 'plugins_loaded', __NAMESPACE__ . '\replace_core_ui_with_custom' ); // Must run after Two Factor plugin loaded.
 add_action( 'init', __NAMESPACE__ . '\register_block' );
+add_action( 'enqueue_block_assets', __NAMESPACE__ . '\maybe_dequeue_stylesheet', 40 );
 
 /**
  * Registers the block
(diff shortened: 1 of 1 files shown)


adamwoodnz:
These styles are still being loaded on Showcase and affecting our performance scores

Screenshot 2023-10-10 at 4 11 18 PM

They're also being loaded on the WordPress.org homepage. Could we use the URI to determine whether to dequeue?

```php
function maybe_dequeue_stylesheet() {
        // Allow to enqueue on the user settings page.
	if ( 0 === strpos( $_SERVER['REQUEST_URI'], '/support/users' ) ) {
		return;
	}

	wp_dequeue_style( 'wporg-two-factor-settings-style' );
}
```

iandunn:
🤔 That's a good point. Fixed in #262 👍🏻

PR #274, opened 2024-07-01
Fix tests

dd32:
I don't fully understand all the changes I've made in this PR, but it makes things pass.


adamwoodnz:
I had to update the .nvmrc to Node 18 to get an install to work

```
yarn install v1.22.19
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error @wordpress/base-styles@5.2.0: The engine "node" is incompatible with this module. Expected version ">=18.12.0". Got "14.19.3"
error Found incompatible module.
```

package.json
```diff
@@ -8,8 +8,8 @@
   "private": true,
   "scripts": {
     "wp-env": "wp-env",
-    "setup:tools": "yarn wp-env run composer install",
-    "test": "yarn wp-env run phpunit phpunit -c /var/www/html/wp-content/plugins/wporg-two-factor/phpunit.xml.dist",
+    "setup:tools": "yarn wp-env run cli --env-cwd=wp-content/plugins/wporg-two-factor composer install",
```
adamwoodnz, line 11: I missed this and didn't run it, just `composer install && yarn test` worked for me

dd32: I think this depends on the version you have installed locally; I found running composer via the containers was more reliable.

The composer is also expecting to be able to modify `../../../mu-plugins/pub` which won't be in the right place when run outside of the containers.

dd32:
> I had to update the .nvmrc to Node 18 to get an install to work

I'm running v20 locally, but evidently whatever is default in github actions is working.. so.. I guess `.nvmrc` isn't being used?

package.json
```diff
@@ -24,8 +24,8 @@
   },
   "homepage": "https://github.com/WordPress/wporg-two-factor#readme",
   "dependencies": {
-    "@wordpress/base-styles": "^4.17.0",
-    "@wordpress/env": "^10.1.0"
+    "@wordpress/base-styles": "^5.2.0",
+    "@wordpress/env": "^10.2.0"
```
adamwoodnz: And the tests pass for me with just the babel and jest configs, these updates don't seem to be required, but maybe there's value in them.

adamwoodnz:
> > I had to update the .nvmrc to Node 18 to get an install to work
> 
> I'm running v20 locally, but evidently whatever is default in github actions is working.. so.. I guess `.nvmrc` isn't being used?

Yeah the workflows aren't calling `nvm use` which isn't ideal.

dd32:
I've bumped the nvm version, and pulled the package updates (since I didn't verify if it didn't change anything).

I broke my local environment by restarting wp-env, so pushed it to see if the actions pass or fail..

package.json
dd32: Reverted them, since I didn't test.

adamwoodnz (approved):
LGTM

PR #278, opened 2024-07-04
Append github hash to block.json version field.

dd32:


StevenDufresne:
Thanks for this. I was going to come back for it.

PR #280, opened 2024-07-04
Add a SVN Password functionality

### LINKED ISSUE #276: Add a way to define an svn password
As 2FA can't be used in all systems (for example, HTTP Basic Authentication) we need a way for users to be able to generate an application password for use with those systems.

This should be a password that is generated by us, is long, contains no special characters that would cause issues with escaping, and be non-memorable / non-guessable.

This password should be able to be re-generated at will, at which time, the other instances of the password become invalid.

For an example of what such a password could look like, with 64 `[A-Za-z0-9]` characters.
```
KdkcGr8yB6ZqUjCNJXVE6x4ujNgviERd8HLyvVoYyPam2NKE2gJGoGE8x2ZcPLZw
```

Additionally the implementation of the SVN passwords should be using a method that allows us to join the Github secret-scanning service; https://docs.github.com/en/code-security/secret-scanning/secret-scanning-partnership-program/secret-scanning-partner-program 

That would require the password be in the format of something like..
```
svn_CRKqyoKl1crIJkgmWfmJDvaeYR98nsJcGicNH89c2VLaiIpWb2086abe
```

dd32:
This is a user_pass but specific for SVN, in it's own table.

This is in place of an application password, fixes #276.

The option only shows when the user requires SVN access (or has a password set), based on user access.

| Overview, not set. | Overview, SVN password set |
| --- | --- |
| Screenshot 2024-07-04 at 4 41 29 PM | Screenshot 2024-07-04 at 4 41 53 PM

| Screen | After password is generated |
| --- | --- |
| Screenshot 2024-07-04 at 4 41 37 PM |Screenshot 2024-07-04 at 4 41 46 PM |

TODO:
 - [x] Verify table structure and names
 - [x] Add user flags when a user is a comitter, plugin author, theme author, etc.
 - [x] Wording.


settings/rest-api.php
```diff
@@ -131,6 +132,33 @@ function register_rest_routes() : void {
...
+			},
+			'args' => array(
+				'user_id' => array(
+					'required' => true,
+					'type'     => 'number',
+					'sanitize_callback' => 'absint',
+					'validate_callback' => function( $user_id ) {
+						return get_userdata( $user_id ) instanceof WP_User;
+					},
+				),
```
xknown, line 164: Do we allow user impersonation in WP.org? If not, maybe we can just use the current user id?

settings/rest-api.php
```diff
@@ -131,6 +132,33 @@ function register_rest_routes() : void {
...
+		array(
+			'methods'  => WP_REST_Server::EDITABLE,
+			'callback' => function( $request ) {
+				$user = get_userdata( $request['user_id'] );
+				return [
+					'svn_password' => set_svn_password( $user->ID )
+				];
+			},
+			'permission_callback' => function( $request ) {
+				return Two_Factor_Core::rest_api_can_edit_user_and_update_two_factor_options( $request['user_id'] );
```
xknown, line 154: Should we restrict this endpoint to users only need a svn password? It seems like any authenticated .org user will be able to set their own SVN password here regardless if they need it or not.

settings/rest-api.php
```diff
@@ -323,6 +351,51 @@ function register_user_fields(): void {
...
+				}
+
+				return false;
+			},
+			'schema' => [
+				'type'    => 'boolean',
+				'context' => [ 'edit' ],
+			]
+		]
+	);
```
xknown, line 389: Is this information exposed for unauthenticated users via the `/wp/v2/users/*` endpoints? If so, should we expose them?

settings/rest-api.php
dd32: This was intentional, as WordPress is not aware of every user that may need a SVN password, such as those who are granted directly via svn configuration files (ie. develop.svn.wordpress.org).

Leaving this open allows for direct links to the svn password screen to be used, rather than having to set some user-meta to allow access.

settings/rest-api.php
dd32: We do, although not how you may have invisioned it here.

Currently a super-admin can modify another users account data, that's why `user_id` is passed to the various 2FA endpoints.

I'm not a super-fan of it, but it's how the other endpoints in the plugin operate, so it made sense to duplicate that here.

An example of where this IS needed, is a case of accounts that cannot be logged into, system-level automation accounts that we need a SVN password for.  However, we do have the CLI commands that can be used for that.

It also allows for super-admins to be able to quickly re-generate a new svn password for a user to lock them out (although, that's not something we'd likely make use of).

Happy to change this to throw a error along the lines of `You can only re-generate a password for yourself` if you feel strongly.

settings/rest-api.php
dd32: No, this is only exposed in the `edit` context, which is limited to authenticated users who can edit that given user; which is limited to the actual user and administrators on the given site.

settings/src/components/account-status.js
```diff
@@ -36,6 +41,19 @@ export default function AccountStatus() {
 				${ backupCodesEnabled ? '' : 'not' }
 				verified your backup codes for two-factor authentication.`;
 
+	let svnBodyText = '';
+	if ( svnPasswordRequired || svnPasswordSet ) {
+		if ( svnPasswordRequired ) {
+			svnBodyText = svnPasswordSet
+				? 'You have an active SVN password set. You can request a new one at any time.'
```
tellyworth: Only reviewing wording here:
* Do we need to explicitly define like "SVN (Subversion)"?
* Could be briefer and more direct, like "A SVN password is set" (or active)
* Makes me wonder if there's such a thing as an inactive SVN password?

settings/src/components/account-status.js
```diff
@@ -36,6 +41,19 @@ export default function AccountStatus() {
...
+				: 'You have not set up an SVN password. This is required for SVN access to WordPress.org.';
+		} else {
+			svnBodyText = svnPasswordSet
+				? 'You have an active SVN password set. You can request a new one at any time.'
+				: 'You do not currently require WordPress.org SVN access.';
```
tellyworth: "You not currently require.." seems like a slightly puzzsing inclusion. Should we just not show it at all? Or grey it out?

I assume it's here for users who don't have commit access to anything.

settings/src/components/screen-navigation.js
```diff
@@ -32,7 +32,8 @@ const ScreenNavigation = ( { screen, children } ) => (
 				{ screen
 					.replace( '-', ' ' )
 					.replace( 'totp', 'Two-Factor Authentication' )
-					.replace( 'webauthn', 'Two-Factor Security Key' ) }
+					.replace( 'webauthn', 'Two-Factor Security Key' )
```
tellyworth: Is 'Two-Factor Security Key' the usual term? Should it mention Hardware or Physical Key in the name?

settings/src/components/svn-password.js
```diff
@@ -0,0 +1,106 @@
...
+			setError( apiFetchError );
+		}
+	} );
+
+	const handleCopy = useCallback( () => {
+		try {
+			navigator.clipboard.writeText( generatedPassword );
+			setGlobalNotice( 'Copied to clipboard' );
+		} catch ( error ) {
+			setGlobalNotice( "Couldn't write to clipboard" );
```
tellyworth: Is it possible to detect whether writing to the clipboard is possible before offering a button that won't work?

I can't come up with a neat way of phrasing it off the top of my head, but we could add words to the effect of "you're going to have to type it yourself" (ie offer a solution).

settings/src/components/svn-password.js
```diff
@@ -0,0 +1,106 @@
...
+	}, [ generatedPassword ] );
+
+	const regenerateButtonText = userRecord.record.svn_password
+		? 'Regenerate password'
+		: 'Request password';
+
+	return (
+		<>
+			<p>
+				Your SVN password can be used to commit to WordPress.org SVN repositories, such as
```
tellyworth: Rather than "can be used" (which implies other possibilities), what about "is used" or "is required".

settings/src/components/svn-password.js
```diff
@@ -0,0 +1,106 @@
...
+				for a plugin or theme.
+			</p>
+			<p>
+				If you forget your SVN password, you can generate a new one here. All previous SVN
```
tellyworth: Simpler to say "..you can reset it here"? And warn about invalidating old ones as part of the process.

tellyworth (commented):
Just suggestions/discussion around wording and UX only.

settings/src/components/svn-password.js
tellyworth: On googling: seems like there's no reliable way to detect whether or not it will work.

settings/src/components/svn-password.js
```diff
@@ -0,0 +1,106 @@
...
+			navigator.clipboard.writeText( generatedPassword );
+			setGlobalNotice( 'Copied to clipboard' );
+		} catch ( error ) {
+			setGlobalNotice( "Couldn't write to clipboard" );
+		}
+	}, [ generatedPassword ] );
+
+	const regenerateButtonText = userRecord.record.svn_password
+		? 'Regenerate password'
+		: 'Request password';
```
StevenDufresne: Should this be simply `Generate password`? `Request password` feels like there could be delay in me receiving the password or some type of application process to complete.

settings/src/components/svn-password.js
```diff
@@ -0,0 +1,106 @@
...
+					/>
+				</Notice>
+			) }
+
+			<p className="wporg-2fa__submit-actions">
+				<Button variant="primary" onClick={ handleGenerate } disabled={ isGenerating }>
+					{ isGenerating ? (
+						<>
+							<Spinner />
+							Requesting..
```
StevenDufresne: If we [agree](https://github.com/WordPress/wporg-two-factor/pull/280/files#r1703556341) on "Generate password", this should be "Generating...".

StevenDufresne:
Regarding terminology, "SVN Password" makes me think that as a plugin committer, my wp.org account and SVN commit passwords are different. 

Provided this isn't the only way to commit in the future (which I'm pretty sure it isn't), have we considered other options like:
- Application Password 
- Application Token 
- Application Key
- Deployment Password 
- Deployment Token
- Deployment Key
- SVN Token 
- SVN Key

I think avoiding "password" altogether would make for simpler documentation.

settings/src/components/account-status.js
dd32: I kind of hope that those who need SVN access would know what SVN is, but, it's easy so why not?

> Makes me wonder if there's such a thing as an inactive SVN password?

There isn't, currently. An 'inactive SVN password' would be an account that doesn't have a svn password set, or that we expired it (some time in the future).

settings/src/components/screen-navigation.js
dd32: This is the existing text, and I agree a little funky. The text is the heading of the sub-screens, ie. click into the security keys section and that's this code.

settings/src/components/svn-password.js
dd32: I thought I'd copied that from somewhere else in the project, but it looks like the only other place that has this is the backup codes that was recently reworked.

https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/copy-to-clipboard-button.js 

We can probably do as that does and just not check..

settings/src/components/account-status.js
dd32: There's 2 main states:
 1. The section is not shown
 2. The section is shown (Known core committer; plugin/theme author; **or has a SVN password set**; doesn't automatically trigger for things like.. buddypress commit)

The problem is that we don't actually know 100% when a user does or doesn't need a SVN password; we know probably 99.5% of the time.

Those then break down into:
 3. The user has a SVN password set (which is NOT the user_pass)
 4. The user doesn't yet have a SVN pass set.

Realistically; I think the text can be clarified significantly.

settings/src/components/svn-password.js
dd32:
```suggestion
				Your SVN password is required to commit to WordPress.org SVN repositories, such as
```

dd32:
> Regarding terminology, "SVN Password" makes me think that as a plugin committer, my wp.org account and SVN commit passwords are different.

That's the intention here, they **will be**.

> Provided this isn't the only way to commit in the future (which I'm pretty sure it isn't), have we considered other options like:

This will be the only manner to authenticate against SVN. Their account password will cease to work.

This specifically does not call it a 'Application Password' as it cannot be used elsewhere in WordPress where an Application password can be used. I didn't call it a token or key because SVN asks for a Password (Since it only does Username + Password auth).

settings/src/components/account-status.js
dd32: Noting the text was simplified significantly.

> Do we need to explicitly define like "SVN (Subversion)"?

I wonder if people know it more as `SVN` or `Subversion`.. I actually suspect it's more the former in the WordPress sphere.

looking at other WordPress documentation:
 - https://codex.wordpress.org/Using_Subversion doesn't call it SVN, but refers to our SVN repo's.
 - https://make.wordpress.org/core/handbook/contribute/svn/ refers to 'Subversion' twice, and SVN a bunch more.

I've experimented with `<abbr>` and `Subversion (SVN)`, and have settled on using SVN everywhere but calling it `Subversion (SVN)` in the first sentence on the page. Additionally, linking to the plugins handbook, see [98fcc52](https://github.com/WordPress/wporg-two-factor/pull/280/commits/98fcc52adf7e0e6c9e88d95168d3f89109268869)

dd32:
I noticed that we're no longer showing the username on the profile page, which is problematic for SVN as it uses a case-sensitive username..  so I've shoehorned that into this text too.

Latest iteration:

Screenshot 2024-08-22 at 5 50 58 PM

dd32:
Also a note; If you test this UI and regenerate the password multiple times, the Copy button stays present in the notice when the password vanishes.  Conditionally showing the copy button component based on a conditional caused react to react badly.. in my experience..

StevenDufresne:
> I noticed that we're no longer showing the username on the profile page, which is problematic for SVN as it uses a case-sensitive username.. so I've shoehorned that into this text too.

You are right. It's also mentioned [here](https://developer.wordpress.org/plugins/wordpress-org/how-to-use-subversion/#your-account). Should we re-add it to the profile page?

dd32:
> You are right. It's also mentioned [here](https://developer.wordpress.org/plugins/wordpress-org/how-to-use-subversion/#your-account). Should we re-add it to the profile page?

I'm thinking since it's only relevant to SVN, we can probably keep it on the SVN page.

That's lead me to questioning the design and layout of the screen, how about something like this instead?

| | |
| --- | --- |
| Screenshot 2024-08-23 at 12 08 29 PM |  Screenshot 2024-08-23 at 12 09 07 PM |

settings/src/components/svn-password.scss
```diff
@@ -0,0 +1,5 @@
+.wporg-2fa__svn-password {
+	.wporg-2fa__svn-copy-password {
+		cursor: pointer;
+	}
+}
```
StevenDufresne, line 26: Can we add some simple `code` styling?
```
code {
        display: inline-block;
        padding: 0 3px;
        background: var(--wp--preset--color--light-grey-2, #f6f6f6);
        border-radius: 2px;
}

File: settings/src/components/svn-password.js [hunk 13]
```diff
@@ -0,0 +1,102 @@
...
+				>
+					<Icon icon={ check } />
+					Your new SVN Password:{ ' ' }
+					<code>{ isGenerating ? <Spinner /> : generatedPassword }</code>
+					<CopyToClipboardButton codes={ generatedPassword } />
+				</Notice>
+			) }
+
+			<p className="wporg-2fa__submit-actions">
+				<Button variant="primary" onClick={ handleGenerate } disabled={ isGenerating }>
```
StevenDufresne: Can we use the `isBusy` prop to make it consistent with other buttons?

| Current | Example |
|--------|--------|
| Screen Capture on 2024-08-23 at 11-21-01 | Screen Capture on 2024-08-23 at 11-23-22 (1) |

settings/src/components/svn-password.js
```diff
@@ -0,0 +1,102 @@
...
+					WordPress.org Plugin Developer Handbook
+				</a>
+				.
+			</p>
+
+			<p>
+				Your username for SVN is the same as your WordPress.org account login, but is
+				case-sensitive. When using SVN you will need to use the username{ ' ' }
+				<code>{ userRecord.record.username }</code>. For security, your WordPress.org
+				account password cannot be used to commit to SVN, you must use a separate SVN
```
StevenDufresne: Since we're doing a soft launch..

```suggestion
				account password should not be used to commit to SVN, use a separate SVN
```

StevenDufresne:
> I'm thinking since it's only relevant to SVN, we can probably keep it on the SVN page.
>
> That's lead me to questioning the design and layout of the screen, how about something like this instead?

Yeah I think those are good additions, letting the user know that there was a password configured in this view which is currently missing. Here's what I came up with based on your designs using some of the existing UI patterns. I think we can remove some of the copy when a password is configured. 

Screenshot 2024-08-23 at 11 49 25 AM

Screenshot 2024-08-23 at 12 03 15 PM

settings/src/components/svn-password.js
dd32: I guess these aren't translated, so we can change it at will. Makes sense.

dd32:
I don't personally see the harm in leaving some of the basic boilerplate text there, how's this?

Current status:

https://github.com/user-attachments/assets/4fe96a3e-5825-486d-957b-62776ce5dece

settings/src/components/svn-password.scss
dd32: I added this to the SVN passwords styles, I guess it's probably worth adding at a higher level though..

settings/src/components/svn-password.js
```diff
@@ -0,0 +1,114 @@
...
+			await refreshRecord( userRecord );
+		} catch ( apiFetchError ) {
+			setError( apiFetchError );
+		}
+	} );
+
+	return (
+		<>
+			<p>
+				WordPress.org uses Subversion (SVN) for version control, every Plugin and Theme
```
StevenDufresne: Can we shorten this to avoid using WordPress.org and SVN twice? It makes it a bit hard on the eyes.

```
WordPress.org uses Subversion (SVN) for version control, providing each hosted plugin and theme with a repository that the author can commit to.
```


| Before | After |
|--------|--------|
| Screenshot 2024-08-28 at 3 06 01 PM | Screenshot 2024-08-28 at 3 07 12 PM |

dd32: LGTM

settings/src/components/svn-password.js
```diff
@@ -0,0 +1,114 @@
...
+
+			<h3>Details</h3>
+			<ul>
+				<li>
+					Username: <code>{ userRecord.record.username }</code>{ ' ' }
+					{ userRecord.record.username.match( /[^a-z0-9]/ ) && <>(case-sensitive)</> }
+				</li>
+				<li>
+					Password:{ ' ' }
+					{ generatedPassword || userRecord.record.svn_password_created ? (
```
StevenDufresne, line 89: Is the `generatedPassword` check necessary? wouldn't `userRecord.record.svn_password_created` be set if a password was generated?

settings/rest-api.php
```diff
@@ -323,6 +357,66 @@ function register_user_fields(): void {
...
+
+	register_rest_field(
+		'user',
+		'svn_password_created',
+		[
+			'get_callback' => function( $user ) {
+				// Local environment doesn't have the SVN password system, just return false for that.
+				if ( ! function_exists( 'WordPressdotorg\Security\SVNPasswords\get_svn_password_creation_date' ) ) {
+					return false;
```
StevenDufresne, line 398: Since this returns `false`, it doesn't allow us to test locally because of [this check](https://github.com/WordPress/wporg-two-factor/pull/280/files#diff-e0a1347d34330bff3a6fb9582b587a9cef40122e84ad844ae7e290311d5c0dbcR77). Can we return a hard-coded date for testing?

dd32: Yep, but not really at the same time.

You can return Dummy data here, but if you hard-code dummy data you can't then test the branch where you've not set a password.

The best option would probably be to add a local test variant of the svn password functionality that uses user meta.

settings/src/components/svn-password.js
dd32: Technically you're probably right, it might not be _required_, since we set `userRecord.record.svn_password_created` above manually. 

In practise, I found it to be required, as without it it caused the UI to "flash" as it re-rendered the branch.

StevenDufresne:
@dd32 I've added some commits, mostly style, language updates. Can you verify that they are ok. 

The last thing on my mind is whether we should change the screen from "SVN Password" to "SVN Credentials".

dd32:
+1 for `SVN Credentials`.

The copy changes appear good to me. I haven't re-tested it yet.

Noting that the systems change has been made, so this can now be used in production.

PR #284, opened 2024-07-10
Add copy/print/download buttons to code backup.

### LINKED ISSUE #45: Add  copy/print/download buttons to Backup Codes
These are in the design in #18, but were [skipped in the MVP milestone to cut scope](https://github.com/WordPress/wporg-two-factor/issues/39#issuecomment-1398555494). We should add them during the first iteration milestone.

StevenDufresne:
Fixes: https://github.com/WordPress/wporg-two-factor/issues/45

This PR adds a copy, print, and download button to the backup code page.

The only downside is that `@wordpress/icons` doesn't have a print icon and we get a weird layout if we combine `<Icon>` with `<Dashicon>`, so I've started with using dashicons.

@WordPress/meta-design Do we have a print `<svg>` that I can use?

## Screenshot

Screenshot 2024-07-10 at 1 39 53 PM

### DIFF
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -3,7 +3,7 @@
  */
 import apiFetch from '@wordpress/api-fetch';
 import { useContext, useCallback, useEffect, useState } from '@wordpress/element';
-import { Button, CheckboxControl, Notice, Spinner } from '@wordpress/components';
+import { Button, ButtonGroup, CheckboxControl, Flex, Notice, Spinner } from '@wordpress/components';
 import { Icon, warning, cancelCircleFilled } from '@wordpress/icons';
 
 /**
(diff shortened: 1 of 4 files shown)


jasmussen:
There should be both copy and download icons, and I'm happy to help with a print icon. However in this case, the action item row here feels like it could easily afford giving slightly more prominence to those buttons, so you could simply spell out copy, print, download as labels.

Should it also be "Done" instead of "All Finished"? Otherwise, probably sentence case, "All finished".

StevenDufresne:
> Should it also be "Done" instead of "All Finished"? Otherwise, probably sentence case, "All finished".

Yeah, that's a good update. I don't know where that copy came from originally.

StevenDufresne:
Alright, I replaced the icons with text.

dd32 (approved):
👍

PR #286, opened 2024-07-15
Add a wizard component for first time configurations.

StevenDufresne:
It's likely that we enforce 2fa for plugins/theme committers on wp.org. This PR adds a wizard that we can show to users on login that will immediately get them into the configuration flow.

## Main changes
- Moved current `script.js` code into `settings.js` so we can choose different UIs.
- Add `onSuccess` events to components to be used in the wizard.
- Rename "account-status" to "home" so we can use the same onSuccess functions but define different 'home' controls.
- Move the unused Progress bar into the 'first-time' folder.
- Allow configuration of "ScreenNavigation" title.

# Screenshots

Screenshot 2024-08-12 at 2 05 17 PM
<hr>

Screenshot 2024-08-12 at 2 05 23 PM

<hr>

Screenshot 2024-08-12 at 2 06 00 PM

<hr>
Screenshot 2024-08-12 at 2 06 15 PM

## How to test
Without local modifications to the underlying WP 2fa plugins, you can't test this. This is best tested in a sandbox.

1. Apply code changes.
2. Navigate to https://profiles.wordpress.org/me/profile/edit/group/3/?first-time=true
3. Try configuring 2fa

Note, if you use an existing account, the wizard will show the "configured states" which isn't expected for regular users.


adamwoodnz:
Using the tab key I can still navigate around behind the wizard. Probably need to trap the tab navigation within.

adamwoodnz:
Generally we use sentence case for titles, and I can see a lot of title case. The bit that sticks out for me is the title below the logo.



Overall I think this could do with some design input; as well as the copy the progress bar details could be refined a little (line thickness, icons).

adamwoodnz:
These back arrows are still using the old blue

Screenshot 2024-08-12 at 12 22 14 PM

StevenDufresne:
> Generally we use sentence case for titles, and I can see a lot of title case. The bit that sticks out for me is the title below the logo.

I've lowercased that option titles. The main titles are existing and enforced using CSS. I think we can address that later if needed.

StevenDufresne:
>  These back arrows are still using the old blue

I've updated this here: https://github.com/WordPress/wporg-two-factor/commit/06f3e14dd2db2f51dd0cc724788f62ec5b5fb82e

settings/src/components/first-time/first-time.js
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * Render the correct component based on the URL.
+ */
+export default function FirstTime() {
+	const modalRef = useRef( null );
+	const { navigateToScreen, screen } = useContext( GlobalContext );
+	const steps = [
+		{
+			title: 'Choose an authentication method',
```
adamwoodnz: Do these strings need to be translatable?

settings/src/components/first-time/first-time.js
```diff
@@ -0,0 +1,212 @@
...
+									window.location.href =
+										'//profiles.wordpress.org/me/profile/edit/group/1';
+								}
+							} }
+							isPrimary
+						>
+							Continue
+						</Button>
+					</div>
+				</div>
```
adamwoodnz: Any reason not to extract this as a component too?

settings/src/components/first-time/first-time.js
```diff
@@ -0,0 +1,212 @@
...
+			),
+		},
+	};
+
+	// Lock the scroll when the modal is open.
```
adamwoodnz:
```suggestion
	// Lock the scroll when the modal is open, and trap tab navigation.
```

settings/src/components/first-time/first-time.js
```diff
@@ -0,0 +1,212 @@
...
+		// Focus the first focusable element in the modal when it opens
+		firstFocusableElement.focus();
+
+		return () => {
+			modal.removeEventListener( 'keydown', trapFocus );
+			document.querySelector( 'html' ).style.overflow = 'initial';
+		};
+	}, [ screen ] );
+
+	const currentStepIndex = screens[ screen ].stepIndex;
```
adamwoodnz, line 122: We could do with a check here for the screen existing. Between switching from prod to sandbox I ended up on `?first-time=true&screen=account-status`, resulting in an error. Shouldn't be possible normally of course but it highlighted an edge case.

settings/src/components/first-time/setup-progress-bar.js
```diff
@@ -0,0 +1,48 @@
...
+
+		if ( currentStepIndex > index ) {
+			return 'is-complete';
+		}
+
+		return 'is-disabled';
+	};
+
+	return (
+		<div className="wporg-2fa__progress-bar">
```
adamwoodnz, line 25: We could likely [make this element accessible](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role) without much effort.

settings/src/components/first-time/first-time.js
StevenDufresne: Not available for this project yet. 

It's logged: https://github.com/WordPress/wporg-two-factor/issues/273

adamwoodnz:
Backup codes print view could do with some elements being hidden

adamwoodnz:
This might be an existing issue but the back button and title don't fit

adamwoodnz:
And the security key screen is collapsing

adamwoodnz (commented):
Flow works well! Feedback is mostly design related or enhancements. Only the mobile styling is likely blocking.

adamwoodnz:
The outline focus style on these radio elements should be removed imo. The focus ring around the whole option and color change on the input should be enough.

jasmussen:
Nice, looks good! Appreciate you using the new typography, colors, and componentry to build this out.

Overall, I wouldn't block this on any one particular piece, but a few things could be refined a bit. In no particular order:

Screenshot 2024-08-12 at 12 42 58

We should establish and adhere to a UI text-case guide. I _think_ one exists, but the link to it escapes me. If I'm not wrong, it's _sentence case with capitals for only proper nouns_. In this case, it would be "Choose an authentication method", and probably "Configure two-factor", unless "two-factor" counts as a proper noun?

Screenshot 2024-08-12 at 12 45 43

Are the icons scaled down? With few exceptions we should only use those icons in 24x24px. In this case you might also want to replace the icons with numbers, 1, 2, 3 instead, as the icons themselves are not perfect. 

Can you apply a 1.5px stroke-width both for the horizontal bar, and for the inner stroke of the circle?

Screenshot 2024-08-12 at 12 49 19

We can refine the Back button a bit: no underline, 24x24 small-chevron icon, same blueberry color. This is similar to what we did for [older 2fa mockups](https://www.figma.com/design/6KPZbXujMGDRWWDR5odjDE/WP.org-improvements-%2F-QA?node-id=565-11747&t=ltmdwwW8CWT0y8SH-1).

security key

Screenshot 2024-08-12 at 12 50 50

Do we have yet a new notice component based on [these designs](https://www.figma.com/design/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?node-id=1264-4116&t=q0WhH9kOcHcfvCtC-1) from @fcoveram?

Screenshot 2024-08-12 at 12 51 44

If not, I would simply make the text bold and not box it up in that yelllow notice piece. It's an old icon, and old colors, and there's a checkbox below to make sure you understand.

Screenshot 2024-08-12 at 12 52 49

Can these buttons be right below the codes themseles?

Screenshot 2024-08-12 at 12 53 48

Instead of "All Finished", should it be "Finish" or "Complete", or even "I understand"? Can it be a disabled solid button instead of a disabled outline button? I.e. this:

Screenshot 2024-08-12 at 12 55 04

Screenshot 2024-08-12 at 12 55 13

Can we omit "Setup Complete" here? Since there's no background, the centered heading becomes a bit off-balanced given there's also a heading right below.

I'd also tend to omit the emoji here. If we want something celebratory, we could potentially revisit the animation we explored for the past mockups:

animation

StevenDufresne:
> Do we have yet a new notice component based on [these designs](https://www.figma.com/design/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?node-id=1264-4116&t=q0WhH9kOcHcfvCtC-1) from @fcoveram?

This plugin was built using `wordpress/components` so we don't have access to wp.org's notices. We could manually restyle it. 

I wonder if we can remove it in favor of hiding the "back" button when codes are generated that need saving and updating the copy to include some of this language?

StevenDufresne:
> Can it be a disabled solid button instead of a disabled outline button? I.e. this:

It can, but if we do, lets do that for all the other buttons as well. Can we open a new ticket for that?

StevenDufresne:
> Can we omit "Setup Complete" here? Since there's no background, the centered heading becomes a bit off-balanced given there's also a heading right below.

It's probably easier to remove in the header in the content than re-working the component title. Its probably not as important.

jasmussen:
> It can, but if we do, lets do that for all the other buttons as well. Can we open a new ticket for that?

Just to understand, the button component by default has an outline-look when disabled, regardless of its base variant? If yes, I'll open an issue, I'd expect primary buttons to remain primary-looking when disabled.

StevenDufresne:
> Just to understand, the button component by default has an outline-look when disabled, regardless of its base variant? If yes, I'll open an issue, I'd expect primary buttons to remain primary-looking when disabled.

We have a few use cases in different components where the button defaults to disabled, we'll want to keep those consistent whether the button is outlined or filled in.

jasmussen:
Exactly, if it's filled, it should always be filled (and have a disabled state), etc. It sounds like that's the case here, so why's it outlined if it's primary when not-disabled? Forgive me if I still miss nuance.

Past experience tells me the best disabled state is just an opacity change (e.g. `opacity: 0.3`), rather than a bespoke visual style.

StevenDufresne:
Closing, we'll go with an inline version: https://github.com/WordPress/wporg-two-factor/pull/297

PR #287, opened 2024-07-15
Fix confirming email  on profiles

dd32:
Since the Account screen was moved from bbPress /  support to Profiles / BuddyPress we've lost the bbPress front-end email confirmation flow.

This restores email changing ability for those who can't access wp-admin.

### DIFF
--- a/settings/rest-api.php
+++ b/settings/rest-api.php
@@ -411,20 +411,20 @@ function require_email_confirmation( $insert_data, $request ) {
  * Process the email change confirmation click.
  */
 function process_email_change_confirmation() {
-	global $current_user;
-	if ( empty( $_GET['newuseremail'] ) ) {
+	$user_id = bp_displayed_user_id();
+	if ( ! $user_id || empty( $_GET['newuseremail'] ) ) {
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


dd32:
If upon loading a page with the `newuseremail` query var, and the current user has no pending email change, they get this:

Screenshot 2024-07-15 at 4 57 09 PM

Note; that the notice has no validation that an email change actually occurred, only that the request attempted to change the address, and that by virtue of being able to access the account screen, the current user could edit that user.

In the event that the user cannot access that screen, they're presented with the users profile and this obscure error (ie. they're logged into the wrong account).

Screenshot 2024-07-15 at 5 00 09 PM

dd32:
Example of the email the user gets to confirm their email address. The text is unchanged, only the URL has changed:

Screenshot 2024-07-15 at 5 07 59 PM

Worth noting, that the post-email-changed email now refers to "WordPress.org Profiles" rather than "WordPress.org Forums", I think the change to Profiles is actually more relevant than Forums.

| Current | Previous |
| --- | --- |
| Screenshot 2024-07-15 at 5 09 05 PM | Screenshot 2024-07-15 at 5 10 41 PM |

Issue #20, opened 2022-11-28
Log 2FA activities


iandunn:
It'd be good to log Stream or MC events when:

* A 2FA provider is enabled/disabled - including the user who changed it (b/c it could be a super admin)
* A backup code is used
* The recovery process is used (details TBD)
* Anything else?

That way we have an audit trail to help if an account is compromised, and have stats around how often recovery is needed, etc.

Related https://github.com/WordPress/two-factor/pull/462, https://github.com/WordPress/two-factor/issues/476

It may be best to add something like `do_action( 'two_factor_log', $event )` upstream, and that could be used for those issues above and for this issue. Having a centralized action would let this plugin automatically log new events that are added upstream, but we still conditionally ignore them if we wanted to.

pkevan:
> Anything else?

Depending on if we allow it or not:
- when backup codes are regenerated
- when a user runs out of unused backup codes since they are prompted to regenerate via an admin notice

I also wonder if we want to log the password reset due to 2FA failure, as per: https://github.com/WordPress/two-factor/pull/482/

iandunn:
Yeah, those are good ideas 👍🏻

dd32:
I think this would make sense to Contribute upstream to Stream, so I've opened https://github.com/xwp/stream/issues/1386

pkevan:
A lot of the potential logging hooks would move with the introduction of https://github.com/WordPress/two-factor/pull/504, so probably worth waiting until that is merged before attempting to add in any new hooks.

iandunn:
@pkevan `504` has merged now 👍🏻

tellyworth:
Decision needed: is this a blocker for launching the MVP, or should we punt it to the next iteration?

iandunn:
I think we can punt to clearly indicate that it's not a blocker for the MVP. If it merges before we launch, though, then we can definitely include it.

dd32:
I've started some some logging via https://github.com/WordPress/wporg-mu-plugins/pull/643 

Included (as of right now)
 - Backup codes regenerated
 - Provider enabled/disabled
 - Authenticated via 2FA (and which method)
 - TOTP setup
 - Security Key registrations & deletions.

Things not covered (yet?)
 - A backup code is used (Technically, I guess that's inferred by the auth'd by backup method?)
 - when a user runs out of unused backup codes (I don't think this is worth logging on?)
 - the password reset due to 2FA failure (There's ZERO actions to catch this on AFAICT)

[...]

PR #289, opened 2024-07-30
Keep stats of logins by 2FA type.

dd32:
See https://github.com/WordPress/wporg-two-factor/issues/20#issuecomment-2256101255


dd32:
Added and backfilled what I could for the stat from the current user sessions. Doesn't account for users who used logout.

pkevan (commented):
👍

pkevan:
> Added and backfilled what I could for the stat from the current user sessions. Doesn't account for users who used logout.

Was wondering how the stats had historical data - thanks for mentioning!

dd32:
I was going to wait for a review, but then found the partial historical data and just went with it :)

Partial data is better than nothing, and will at least allow a bit more of a baseline for the next month.

Issue #253, opened 2023-09-18
Not redirected to Backup Codes after enabling WebAuthn


iandunn:
1. Wipe out all settings/providers so you start fresh
2. Enable TOTP. You're redirected to Backup Codes
1. Wipe out all settings/providers so you start fresh
1. Enable WebAuthn. You're not redirected to Backup Codes

For consistency, I think we should add something like this to WebAuthn
    https://github.com/WordPress/wporg-two-factor/blob/fa4c81b35d0c6b8fe75723ee2529299a0b2e767e/settings/src/components/totp.js#L25-L28

dd32:
#286 might resolve this.

Alternatively, given you can enable multiple keys, perhaps simply a notice at the bottom of the keys screen suggesting "All done? Set some Backup codes".

This is IMHO one of the reasons why it's common to be presented with the backup codes **prior** to setting up any 2FA methods.

PR #296, opened 2024-08-14
Only create backup codes when user initiated

### LINKED ISSUE #250: Add button to explicitly generate backup codes
If you visit the Backup Codes screen for the first time, the codes are generated automatically, even if the user doesn't intend to (they may just be exploring the feature, and not ready to permanently save the codes, etc). That worsens the effects of https://github.com/WordPress/two-factor/issues/507.

We could add a button as the first step, so we at least know the user intends to _generate_ codes (if not _save_ them). It seems like a [bad practice to take an action that modifies their account without an explicit direction](https://ux.stackexchange.com/questions/6506/automatic-ui-without-explicit-perform-action-controls).

We could provide a way to disable backup codes so they can [undo an accidental generation](https://www.nngroup.com/videos/usability-heuristic-user-control-freedom/), but it's important to have them to prevent getting locked out of their account. If we do that, we should probably disable all 2FA providers, and make it very clear to them that we've done that and they'll need to set it up again.

Related #256

adamwoodnz:
Fixes #250 

Changes the backup codes screen to not generate codes on load. Instead the `Generate codes` button must be activated. Once backup codes have been enabled the button text changes to `Regenerate codes` as before.

Improves the UI by not rendering the action buttons until the codes are available.

| Pre-generation | Enabled |
|--------|--------|
| Screenshot 2024-08-14 at 1 48 57 PM | Screenshot 2024-08-14 at 1 49 27 PM |

### DIFF
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -237,7 +237,9 @@ function Manage( { setGenerating } ) {
 			</div>
 
 			<Button isSecondary onClick={ () => setGenerating( true ) }>
-				{ backupCodesEnabled ? 'Regenerate backup codes' : 'Generate backup codes' }
+				{ backupCodesEnabled
+					? 'Regenerate and save backup codes'
+					: 'Generate and save backup codes' }
 			</Button>
 		</>
 	);
--- a/settings/src/components/backup-codes.js
+++ b/settings/src/components/backup-codes.js
@@ -132,12 +132,6 @@ function Setup( { setGenerating, onSuccess } ) {
 				<>
 					<CodeList codes={ backupCodes } />
 
-					<ButtonGroup>
-						<CopyToClipboardButton contents={ backupCodes } />
-						<PrintButton />
-						<DownloadButton codes={ backupCodes } />
-					</ButtonGroup>
-
 					<CheckboxControl
 						label="I have printed or saved these codes"
 						checked={ hasPrinted }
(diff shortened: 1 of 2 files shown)


settings/src/components/backup-codes.js
```diff
@@ -41,25 +40,20 @@ export default function BackupCodes() {
 		);
 	}
 
-	if (
-		! backupCodesEnabled ||
-		backupCodesRemaining === 0 ||
-		regenerating ||
-		! backupCodesVerified
```
adamwoodnz, line 48: These are the conditions which led to the codes being generated on load

settings/src/components/backup-codes.js
```diff
@@ -134,12 +124,6 @@ function Setup( { setRegenerating } ) {
 				<>
 					<CodeList codes={ backupCodes } />
 
-					<ButtonGroup>
-						<CopyToClipboardButton codes={ backupCodes } />
-						<PrintButton />
-						<DownloadButton codes={ backupCodes } />
-					</ButtonGroup>
```
adamwoodnz: Moved to the CodeList and conditionally rendered

settings/src/components/backup-codes.js
StevenDufresne: We currently auto navigate to the backup code component when a user sets up a one time password. We'll also do that in the onboarding [flow](https://github.com/WordPress/wporg-two-factor/pull/297) and add that support for [Webauthn](https://github.com/WordPress/wporg-two-factor/issues/253). Now that this doesn't auto trigger, do you think we should export the "Setup" component and handle that in those flows? Alternatively we could pass a prop into this component. Thoughts?

adamwoodnz: Ah I wasn't aware that they need to be auto generated in those cases. Yeah passing a prop seems reasonable.

adamwoodnz: Sorry I'm confused about this. We can still auto navigate to backup codes, but my understanding is that we still should not auto generate. Is it a requirement that codes are generated if one of the 2FA methods has been activated? If it is then auto generation still makes sense.

StevenDufresne: It's not a requirement per se, but we want to limit the number of accounts that don't have recovery codes as much as possible seeing that regaining access to a locked out account can be troublesome for users.

adamwoodnz: Yeah I guess it's a balance between acting on their account without their action vs ensuring they don't get locked out. I'll add the auto-generation so we can see what it looks like.

adamwoodnz: Reading through #507, and in particular @dd32's [comment](https://github.com/WordPress/two-factor/issues/507#issuecomment-1425108260), I don't think auto generating is the best path here. If we're going to auto generate it feels like we should only display the codes, and require a second action to save. I've updated the button text in this PR to be more explicit about the action taken if they do choose to generate. What do you think?

settings/src/components/backup-codes.js
```diff
@@ -24,10 +24,9 @@ import DownloadButton from './download-button';
  */
 export default function BackupCodes( { onSuccess = () => {} } ) {
 	const {
-		user: { backupCodesEnabled, hasPrimaryProvider, backupCodesRemaining },
-		backupCodesVerified,
+		user: { hasPrimaryProvider },
```
adamwoodnz, line 27: This was previously undefined. Sometimes I wish we had TS for this sort of thing.

settings/src/components/settings.js
```diff
@@ -20,7 +20,11 @@ import { GlobalContext } from '../script';
  * Render the correct component based on the URL.
  */
 export default function Settings() {
-	const { backupCodesEnabled, navigateToScreen, screen } = useContext( GlobalContext );
+	const {
+		user: { backupCodesEnabled },
```
adamwoodnz, line 25: This was undefined. Sometimes I wish we had TS for things like this.

settings/src/components/backup-codes.js
```diff
@@ -241,8 +236,10 @@ function Manage( { setRegenerating } ) {
 				) }
 			</div>
 
-			<Button isSecondary onClick={ () => setRegenerating( true ) }>
-				Generate new backup codes
+			<Button isSecondary onClick={ () => setGenerating( true ) }>
+				{ backupCodesEnabled
+					? 'Regenerate and save backup codes'
+					: 'Generate and save backup codes' }
```
StevenDufresne, line 242: I understand why we want to add "Generate and **save backup codes**" because that's what technically happens but I don't know that first time users will understand what save means here. I could be wrong.

settings/src/components/backup-codes.js
StevenDufresne: I think that's fair. I just tested this. It does help to establish context. 

We can monitor whether users bail on this view and adjust as necessary.

PR #303, opened 2024-08-23
Move up warning message on backup code screen.

StevenDufresne:
With the copy/print/download buttons moving up below the codes, the banner creates a lot of vision disruption. In addition, when this view is displayed the backups are already set, so we want users to see that banner as early as possible so they don't navigate back out.

| Before | After |
|--------|--------|
| Screenshot 2024-08-22 at 3 46 03 PM | Screenshot 2024-08-22 at 3 45 13 PM |


dd32 (approved):
LGTM

PR #297, opened 2024-08-14
Add onboarding flow for new users.

StevenDufresne:
This builds on https://github.com/WordPress/wporg-two-factor/pull/286 but adds code to make the wizard inline.

## Screenshots

profiles wordpress org_dufresnesteven_profile_security_first-time=1 screen=home (1)

profiles wordpress org_dufresnesteven_profile_security_first-time=1 screen=home (2)

profiles wordpress org_dufresnesteven_profile_security_first-time=1 screen=home (3)

profiles wordpress org_me_profile_security (1)

profiles wordpress org_me_profile_security (2)

Issue #291, opened 2024-08-07
2FA setup process should have a singular entry point


dd32:
Currently the Two-Factor authentication setup flow requires the user to enter specific screens.

As suggested via #286 this would be ideally better as a Setup Wizard for a first-timer.

Further to that however, is that the Profile -> Account & Security screen has 3 different cards entry points, which for a non-technical user may be far more confusing than it could be.

I feel that, like with other sites that use 2FA, we should perhaps have a singular 'Enable 2FA' card, similar to below.

| Current | Proposed Enhancement |
| --- | --- |
| <img width="967" alt="Screenshot 2024-08-07 at 1 26 01 PM" src="https://github.com/user-attachments/assets/aab119b7-240c-4ff5-8325-e543d2874a32"> | <img width="962" alt="Screenshot 2024-08-07 at 1 27 17 PM" src="https://github.com/user-attachments/assets/628917a6-9658-4108-8563-c239b2e54341"> |

PR #305, opened 2024-08-26
Redirect to backup codes for webauthn if none setup.

### LINKED ISSUE #253: Not redirected to Backup Codes after enabling WebAuthn

StevenDufresne:
Fixes: #253

This PR hooks into the `<WebAuthn>` `onKeyAdd` function added in #297 to redirect users to `<BackupCodes>` if they don't have backup codes installed yet.

### DIFF
--- a/settings/src/components/settings.js
+++ b/settings/src/components/settings.js
@@ -38,8 +38,16 @@ export default function Settings() {
 				} }
 			/>
 		),
+		webauthn: (
+			<WebAuthn
+				onKeyAdd={ () => {
+					if ( ! backupCodesEnabled ) {
+						navigateToScreen( 'backup-codes' );
+					}
+				} }
+			/>
+		),
 		'backup-codes': <BackupCodes onSuccess={ () => navigateToScreen( 'home' ) } />,
-		webauthn: <WebAuthn />,
 	};
 
 	const currentScreenComponent =


settings/src/components/settings.js
```diff
@@ -38,8 +38,16 @@ export default function Settings() {
 				} }
 			/>
 		),
+		webauthn: (
+			<WebAuthn
+				onKeyAdd={ () => {
+					if ( ! backupCodesEnabled ) {
+						navigateToScreen( 'backup-codes' );
+					}
```
adamwoodnz, line 46: Should we navigate to home if backup codes are already enabled, like TOTP above?
```suggestion
					if ( ! backupCodesEnabled ) {
						navigateToScreen( 'backup-codes' );
					} else {
						navigateToScreen( 'home' );
					}
```

adamwoodnz: Personal pref: I would keep both of these DRYer by using a ternary for the screen name

```
<WebAuthn onKeyAdd={ () => navigateToScreen( backupCodesEnabled ? 'home' : 'backup-codes' ) } />
```

StevenDufresne: > Personal pref: I would keep both of these DRYer by using a ternary for the screen name

I would typically prefer explicitness (within reason) over the use of ternaries for conciseness, in this case. It's easier to extend and debug.

> Should we navigate to home if backup codes are already enabled, like TOTP above?

I don't think so since we have a useful interface to show the user after adding a key (key list) and users can add multiple keys, whereas with TOTP there really isn't anything left to do.

adamwoodnz: Makes sense 👍

adamwoodnz (approved):
LGTM

dd32:
I'm not a super-fan of automatically redirecting after adding a key, even for TOTP it seems a bit weird.. but I get it, it's the only way at present to remind them to setup recovery codes.. 

I question whether this would be needed if something like #291 was implemented which triggered the #297 onboarding?

StevenDufresne:
> I question whether this would be needed if something like https://github.com/WordPress/wporg-two-factor/issues/291 was implemented which triggered the https://github.com/WordPress/wporg-two-factor/pull/297 onboarding?

I think those are fair points. I don't have a timeline for #291. However, I believe the current approach is a reasonable fallback. Since users cannot disable 2FA codes, adding a key without codes shouldn't lead to any awkward scenarios. Anyone who adds a key without codes will not have previously had them enabled.

dd32 (approved):
Based on discussion, I think this LGTM

Issue #306, opened 2024-08-27
General UI Improvements


StevenDufresne:
These improvements were agreed upon in a QA session done between @jasmussen and @StevenDufresne. I've collected them into a single ticket as the are pretty minor.

# To do
- [x] Make default font size `14px` for copy. Button font-size should be `14px`.
- [x] Make chevron (back button) `24x24`. (a78628754d5f08b17ddadd3aeef216a25663125f)
- [x] Apply `text-wrap: pretty` to content. ([c98b9c1](https://github.com/WordPress/wporg-two-factor/commit/c98b9c105d633245dea4634aeb1cd8376db9bf2e))
- [x] Sentence case for titles. (https://github.com/WordPress/wporg-two-factor/pull/308)
- [x] [Surround key name](https://github.com/WordPress/wporg-two-factor/blob/8c63862156c99cbeecc2627d52683a583e34c07a/settings/src/components/webauthn/register-key.js#L94) with "". (dc21342396b202980e276181a901287682a0c694)
- [x] Update backup code button to read "Done" not "All Finished". (ab8bb10e32f0305be8c2e62043977ca1d9d37c23)

StevenDufresne:
Closing as related issues are complete.

PR #308, opened 2024-08-28
Updates heading text to Sentence case.

StevenDufresne:
See #306.

This PR lowercases the view titles and heading to sentence case to match wp.org.

### DIFF
--- a/settings/src/components/screen-navigation.js
+++ b/settings/src/components/screen-navigation.js
@@ -32,14 +32,7 @@ const ScreenNavigation = ( { screen, children, title = '', canNavigate = true } 
 				/>
 			) }
 
-			<h3>
-				{ title.length
-					? title
-					: screen
-							.replace( '-', ' ' )
-							.replace( 'totp', 'Two-Factor Authentication' )
-							.replace( 'webauthn', 'Two-Factor Security Key' ) }
-			</h3>
+			<h3>{ title }</h3>
 		</CardHeader>
 		<CardBody className={ 'wporg-2fa__' + screen }>{ children }</CardBody>
 	</Card>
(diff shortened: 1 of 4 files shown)


settings/src/components/screen-navigation.js
```diff
@@ -32,14 +32,7 @@ const ScreenNavigation = ( { screen, children, title = '', canNavigate = true }
...
 
-			<h3>
-				{ title.length
-					? title
-					: screen
-							.replace( '-', ' ' )
-							.replace( 'totp', 'Two-Factor Authentication' )
-							.replace( 'webauthn', 'Two-Factor Security Key' ) }
-			</h3>
+			<h3>{ title }</h3>
```
adamwoodnz, line 35: ❤️

adamwoodnz:
I'm still seeing uppercase due to the [CSS](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/screen-navigation.scss#L24)

adamwoodnz:
Also how does one enter the onboarding flow? I've tried `/profile/security`, but `is_onboarding_page()` starts off true then becomes false, so I end up in the normal screens.

StevenDufresne:
> I'm still seeing uppercase due to the [CSS](https://github.com/WordPress/wporg-two-factor/blob/trunk/settings/src/components/screen-navigation.scss#L24)

A good catch, I missed committing that.

StevenDufresne:
> Also how does one enter the onboarding flow? I've tried /profile/security, but is_onboarding_page() starts off true then becomes false, so I end up in the normal screens.

You're probably getting stuck because you have 2fa configured: https://github.com/WordPress/wporg-two-factor/blob/9cac7f8f9cd98a382be303beef544c3bcc159632/settings/settings.php#L233

adamwoodnz (approved):
Conflicts, otherwise LGTM

PR #312, opened 2024-09-02
Update 'You've got' to 'you have'.

StevenDufresne:
Update string `You've got a SVN password configured for your account.` to `You have an SVN password configured for your account.`.

I know it's less punchy but consistent with the other status messages.

cc @dd32


dd32 (approved):
Makes sense to me.

PR #313, opened 2024-09-05
Enable the "please enable 2FA" login nag for plugin + theme authors.

dd32:
It's time to flick on the login nag for Plugin and Theme authors, now that it's been announced:
https://make.wordpress.org/plugins/2024/09/04/upcoming-security-changes-for-plugin-and-theme-authors-on-wordpress-org/

This will trigger the 'enable 2fa please' notice on login.

### DIFF
--- a/wporg-two-factor.php
+++ b/wporg-two-factor.php
@@ -273,19 +273,17 @@ function user_should_2fa( $user ) : bool {
 		return true;
 	}
 
-	/*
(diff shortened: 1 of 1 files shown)


dd32:
@StevenDufresne Do you think we should do this today; or leave it a few days to see if authors naturally enable it?

StevenDufresne:
I think we should do it today. No need to wait.

PR #319, opened 2024-10-22
Make sure cap user is active user.

StevenDufresne:
I think this solves an issue with displaying the `get_enable_2fa_notice` in the support forums. It is being displayed because it's running for every user included in the forum table. Run it only for current user.

### DIFF
--- a/wporg-two-factor.php
+++ b/wporg-two-factor.php
@@ -75,7 +75,7 @@ add_filter( 'two_factor_primary_provider_for_user', __NAMESPACE__ . '\set_primar
 add_filter( 'two_factor_totp_issuer', __NAMESPACE__ . '\set_totp_issuer' );
 add_action( 'set_current_user', __NAMESPACE__ . '\remove_super_admins_until_2fa_enabled', 1 ); // Must run _before_ all other plugins.
 add_action( 'login_redirect', __NAMESPACE__ . '\redirect_to_2fa_settings', 105, 3 ); // After `wporg_remember_where_user_came_from_redirect()`, before `WP_WPorg_SSO::redirect_to_policy_update()`.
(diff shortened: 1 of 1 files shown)


wporg-two-factor.php
```diff
@@ -192,16 +192,18 @@ function remove_super_admins_until_2fa_enabled() : void {
...
  * perform privileged actions on the front end, via the REST API, etc.
  */
-function remove_capabilities_until_2fa_enabled( array $allcaps, array $caps, array $args, WP_User $user ) : array {
-	if ( 0 === $user->ID || ! user_requires_2fa( $user ) ) {
+function remove_capabilities_until_2fa_enabled( array $allcaps, array $caps, array $args ) : array {
+	$current_user = wp_get_current_user();
+
+	if ( 0 === $current_user->ID || ! user_requires_2fa( $current_user ) ) {
 		return $allcaps;
 	}
```
dd32: I think this should respect the passed $user, but the issue is that it shouldn't be performing any actions when the user having it's caps checked doesn't match the current logged in user.

Basically, revert all changes here and instead add a `$user->ID !== get_current_user_id()` to the early return.
```suggestion
function remove_capabilities_until_2fa_enabled( array $allcaps, array $caps, array $args, WP_User $user ) : array {
	if ( 0 === $user->ID || $user->ID !== get_current_user_id() || ! user_requires_2fa( $user ) ) {
		return $allcaps;
	}
```

StevenDufresne: Thanks. I left this in the draft because I see that we use `wp_get_current_user` in other filters but not here. I wasn't familiar with when this filter is applied and was going to look into it and figure out why it was being called within the context of the forum tables. If it's as simple as comparing the current user then cool.

Issue #323, opened 2025-01-28
Support WebAuthN plugin v2.5.1+


dd32:
https://github.com/sjinks/wp-two-factor-provider-webauthn/commit/c0198c8ab79bfcc169034296f57053c25d1cc99f

`Admin::is_webauthn_enabled()` no longer returns truthful on 2.5.3, as the provider class isn't present within the provider list.

I'm not sure why, but in the quick testing:
```
var_dump(  \Two_Factor_Core::get_providers() );
array (size=2)
  'Two_Factor_Totp' => 
    object(WordPressdotorg\Two_Factor\Encrypted_Totp_Provider)[1389]
  'Two_Factor_Backup_Codes' => 
    object(Two_Factor_Backup_Codes)[1391]
```

This causes the 2FA interface to be broken as the XHR endpoints for the plugin aren't loaded.

PR #325, opened 2025-07-10
DB: Only customize the DB schema locations when it's not being run locally

dd32:
This avoids having `table not found` errors in local environments, and ensures that schema updates run more as expected.


dd32:
Somewhat unexpectedly this fixes the failing PHPUnit tests, at least for the unit tests.

That suggests that the change that's needed for #323 is somewhat related to the customised code here.

PR #329, opened 2025-07-29
Change in message

revathim007:


revalidation/index.php
```diff
@@ -117,7 +117,7 @@ function enqueue_assets() {
 		'cookieName' => COOKIE_NAME,
 		'l10n'       => [
 			'title'   => __( 'Two-Factor Authentication', 'wporg' ),
-			'message' => __( 'Please verify your Two-Factor Authentication to continue.', 'wporg' ),
+			'message' => __( 'Please Verify your Two-Factor Authentication to continue.', 'wporg' ),
```
dd32, line 120: The capitalisation of `verify` should be lowercase here I believe. Please is capitalised due to start of sentence, and `Two-Factor Authentication` is a feature name.

revalidation/index.php
```diff
@@ -141,7 +141,7 @@ function set_cookie() {
 	setcookie(
 		COOKIE_NAME,
 		$expires_at,
-		$expires_at - MINUTE_IN_SECONDS, // The cookie will cease to exist to JS at this time.
+		$expires_at - 90, // The cookie will cease to exist to JS at this time.
```
dd32, line 144: Increasing this from 60s to 90s doesn't inherently improve anything, I'm unsure why the change is necesary.

The only time when this would be needed is if the action being acted upon takes longer than 60seconds between submission and when the server starts processing.

This could be an issue for submissions where a large upload is occuring at the same time, however, I believe we have a higher timeout on POST requests here, which negates the need for these 30s.

dd32:
Hello @revathim007,

Thanks for your PR, however without any explanation of why the change is being requested I'm forced to close this PR, noting the above issues.

If you'd like to add further details, we can re-open it if necessary.

PR #333, opened 2026-02-11
Remove workaround for modal closing in Firefox

adamwoodnz:
See https://github.com/WordPress/wporg-two-factor/issues/180

Removed temporary workaround for modal closing behavior in Firefox, now that https://github.com/WordPress/gutenberg/pull/51602 is fixed.

Workaround was added in https://github.com/WordPress/wporg-two-factor/pull/185


dd32 (approved):
This should be safe to merge based on my understanding. We def run a newer Gutenberg version too :)

dd32:
@adamwoodnz Please note: If you want to deploy this to w.org you'll need to test _everything_, as the upstream two-factor has changed a bunch of stuff that might cause 2fa to break on w.org.. I'm not actually sure..

PR #335, opened 2026-02-23
Update dependencies, php versions for unit tests, and misc..

dd32:
This is in an attempt to fix the unit tests for #334  which was resulting in plugins being installed into paths such as `$SLUG.latest-stable`.

In the process I've:
 - Updated all deps.
 - Updated node version to 20
 - Updated unit tests to run on PHP 8.1 + 8.5

Note: This includes a workaround for https://github.com/WordPress/gutenberg/issues/75718


dd32:
Cause is an upstream regression in wp-env: https://github.com/WordPress/gutenberg/issues/75806

It's not clear why pre-package-upgrade this problem still existed though.

This PR is pending upstream being fixed, for both this and https://github.com/WordPress/gutenberg/issues/75718 (yarn resolutions change)

package.json
```diff
@@ -24,12 +24,15 @@
...
+    "@wordpress/base-styles": "^6.16.0",
+    "@wordpress/env": "^11.0.0"
   },
   "workspaces": {
     "packages": [
       "settings"
     ]
+  },
+  "resolutions": {
+    "@wordpress/vips": "1.0.1-next.v.202602200903.0"
```
dd32:
```suggestion
```
Revert, hopefully https://github.com/WordPress/gutenberg/issues/75718 is fixed.

dd32:
> It's not clear why pre-package-upgrade this problem still existed though.

`warning Lockfile has incorrect entry for "@wordpress/env@^10.1.0". Ignoring it.`

It appears to be installing ` @wordpress/env@10.39.0`

